Issue #2, opened 2021-11-22
Create a proper README


Dhaiwat10:
We need a proper readme file that contains some details about what this project is, what the current status is and how can one contribute.

swetshaw:
Hey, I am ready to help.

Dhaiwat10:
@swetshaw thanks! Let me know if you need anything

gilcreque:
Maybe this would help https://readme.so/

PR #6, opened 2021-11-25
Readme update

swetshaw:
Progress made for #2 
Provided a skeleton README in this PR
The following needs to be done in the next commits/PRs
- Need to add Starter codes in the README
- Add more details about how this Web3-UI library is different from other web 2 UI libraries
- Link to documentation
- Build using? Typescript, Chakra, ..?


frankTurtle (approved):
Looks good to me :)

swetshaw:
Thanks @frankTurtle , but I will update the README as and when the product starts taking its shape. The RFC [here](https://forum.developerdao.com/t/rfc-web3-ui-a-web3-specific-ui-library/565) is really useful. Kudos to @Dhaiwat10 for such a well-documented RFC.

Dhaiwat10:
Thanks for creating this @swetshaw! I will also make sure that I drop in any useful info here as the project grows

README.md
```diff
@@ -1,3 +1,14 @@
-# web3-ui
+# Welcome to Web3-UI (The UI library for Web3)
+
+Web3-UI has been developed keeping in mind the specific needs of web3 websites.
+
+## Installation
+
+Web3-UI is available as an [npm package](https://www.npmjs.com/package/web3-ui)
+```npm install web3-ui``` or ```npm i web3-ui```
```
crondinini, line 8: We seem to use `yarn`, so I'd suggest changing this to `yarn install`.

I'd also include information about running storybook:

"To start storybook and see the components already created:"
```
yarn storybook
```

crondinini (commented):
I have an opinion around keeping a PR open but I wanted to preempt that this is coming from a good place or at least that's my intention. Hope it comes across that way. 😄 

Also, my lengthy reasoning is an overkill in this specific case, but I'd love to share anyway because it can be helpful in future occasions. It can also be a good start for a discussion. 

Keeping a PR opened for a long time has a few disadvantages in my experience:
1. It's like having a product permanently in the warehouse, it never actually gets to the people that would benefit from it. In other words, the value/information produced never gets delivered.
2. It's a sure way of getting git conflicts, which will then require extra time from the PR author.
3. It pollutes the Pull Requests section which is not a bad thing yet at this point but as the number of contributions grow, having a lot of PRs open makes it harder for reviewers to know what is actually ready for review. A draft PR is in my opinion an invitation for pre-review around an approach or direction and it quickly (~1 day) follows an actual PR.
4. Approvals get stale and it constantly requires new review.
5. It might lead to 2 people working on the same thing. For example, if anyone wants to add to README, it might end up being the same information written twice since the original change was never incorporated.

Considering all of these reasons I would either: add a bit more info to get to a point where you are comfortable merging this or I'd close the PR until you get to such point. A branch is always available so there's no need to have a PR if it's not ready to be reviewed and merged.

README.md
swetshaw: Thank you so much for the feedback @crondinini. Your points are totally valid, I will update the README with ```yarn storybook``` instruction and other necessary information so that this PR gets merged and closed. 🙌

swetshaw:
Created new PR #34 in replacement of #6

PR #16, opened 2021-11-26
Basic Web3 setup and <ConnectWallet />

### LINKED ISSUE #8: Global `Web3Context`, connect wallet functionality and connect wallet UI
## Checklist
- [ ] Global `Web3Context` which will provide different utilities to your app through the `<Provider />`.  The utilities would enable easy access to information like - whether a wallet is connected or not, the network the wallet is connected to, etc.
- [ ] Utility functions like `requestNetworkChange`, `requestWalletConnection`, etc.
- [ ] Connect wallet buttion and any other UI pieces necessary.

## Notes
Try to re-use existing solutions like `web3modal`.

Dhaiwat10:
<will add the description soon>

Closes #8


.storybook/preview.js
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import { ChakraProvider } from '@chakra-ui/react';
+import { Provider } from '../src/components';
 
 export const decorators = [
   (Story) => (
-    <ChakraProvider>
+    <Provider network='rinkeby'>
```
frankTurtle, line 7: Any thoughts on making this an env variable so we don't need to change it for a prod push?

src/components/ConnectWallet/ConnectWallet.tsx
```diff
@@ -0,0 +1,35 @@
+import { Button, ButtonProps } from '@chakra-ui/button';
+import React from 'react';
+import { useWeb3 } from '../../hooks';
```
frankTurtle, line 3: I'd vote on putting the path aliases in haha!

frankTurtle (approved):
🤙🤙

.storybook/preview.js
Dhaiwat10: This won't get pushed to prod. This is just a Storybook config.

src/components/ConnectWallet/ConnectWallet.tsx
Dhaiwat10: Great suggestion!

Dhaiwat10: We are going to divide the hooks and components into separate packages through a monorepo so that'd probably solve that.

Issue #19, opened 2021-11-27
Review/refactor bundler setup


crondinini:
### What is the problem?

We're currently including 2 bundlers as dev dependencies, `rollup` and `webpack`. 

It seems that storybook requires `webpack`. and that our project is using `rollup`. We should look into using one single bundler.

### Why?

Having more than 1 bundler introduces unnecessary complexity. Bundlers are already complicated in nature to configure and debug, it's something done early in a project and it's hard to modify as the project increases in complexity. Having 2 bundlers could potentially increase that complexity further.

### How?

Whoever picks up this issue should answer the questions:
- Which bundler will work for both the project and storybook?
- If there is more than one option, which one is the best one? Other options to look into `Parcel`, `Snowpack`.

### Expected outcome

- One single bundler in the dev dependencies
- One single configuration in the root of the project
- Both storybook and the react project still compile without a problem

Dhaiwat10:
@with-heart had suggested [preconstruct](https://preconstruct.tools/).

Added it here for reference for whoever picks this one up!

etr2460:
I'll take a look into picking the best option here and will report back with a recommendation!

etr2460:
A summary of conversations and research:

**Webpack**
For Storybook to work, it needs a webpack config. So we'll always have webpack in the repo, but it'll only be used for Storybook, so should be pretty decoupled from the build process for releasing new packages.

Parcel and Snowpack also seem more in the vein of Webpack as being focused on applications, not libraries, so they don't seem like the best fit.

**Rollup**
The status quo for building for release, and seems to do everything we need right now. However, we seem to use very little of the features it supports, and the fact that it bundles all our TS into a single JS file seems less than ideal. One reason using rollup (or a more complex builder/bundler) could be valuable would be if we anticipate a large variety of files being needed beyond TS/TSX (css for example, or others)

**tsc**
If all we have is TS/TSX files, then I think we can use the typescript compiler to build all the static assets we need. This would reduce the number of dependencies in the repo, but would require a bit more customization, and makes adding non TS/TSX files to the package a bit more tricky

**babel**
Similar to `tsc`, we could use babel to transpile all the TS/TSX for the build, and also load a variety of other asset types as needed. Requires a bunch of configuration though, and might not be needed since anyone who uses our library could ensure that the code they output will work on their target.

**preconstruct**
[...]

Issue #18, opened 2021-11-27
Configure publish/release flow


crondinini:
### Overview 

After cleaning up items from the original template, one of the things removed was the publish action alongside all dependencies for semantic release. We should think about what's the strategy we want to follow for releasing and if we want to go the automated route of using a package for the semantic release.

```
  publish:
    if: ${{ github.ref == 'refs/heads/master' }}
    needs: test-lint-build
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          persist-credentials: false
      - uses: actions/setup-node@v1
        with:
          node-version: 12
      - name: Restore node modules cache
        uses: actions/cache@v2
        env:
          cache-name: cache-node-modules
        with:
          path: node_modules
          key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
      - name: Install Dependencies
        run: yarn install
      - name: Build
        run: yarn build
      - name: Release
        env:
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
          GITHUB_TOKEN: ${{ secrets.SEMANTIC_RELEASE_TOKEN }}
        run: npx semantic-release
```

Dhaiwat10:
Whenever we decide to work on this problem, we should also look at creating a workflow for a well-maintained changelog.

PR #20, opened 2021-11-27
chore: clean up deps and upgrade node version

crondinini:
### Overview 

This PR does the following:

- removes a few dev deps, namely
   - css related dependencies that seemed to have been previously used in storybook
   - everything related to publishing and semantic release (instead of that, I created [an issue](https://github.com/Developer-DAO/web3-ui/issues/18) for us to look into that)
      - this included removing an action from Github action 
   - commit lint for conventional commits 
- upgrades the node version used
- reverts adding everything to the stage during the pre-commit hook
- deleted the changelog that was being automatically generated

### Excluded from the PR

There are more things we could look into simplifying that I didn't include here due to my own time constraints:
- Both rollup and webpack are being referenced. It seems that storybook requires webpack. We could look into using only one bundler that works for both things. I've created [an issue](https://github.com/Developer-DAO/web3-ui/issues/19) for that.
- Initially I deleted the whole set up around the templates with hygen, but then I saw commits that indicated to me they were being used.
- I also didn't deleted the `Examples.tsx` but it seems that we could. Why do we need both that and the templates? But I didn't want to create conflicts in other PRs so I left it as is.


.husky/pre-commit
```diff
@@ -2,4 +2,3 @@
 . "$(dirname "$0")/_/husky.sh"
 
 yarn format
-git add .
```
Nazeeh21, line 5: Just curiously asking is there any particular reason for removing this?

Nazeeh21: Overall changes LGTM, though would love to hear from @Dhaiwat10.

Nazeeh21:
Overall changes LGTM, though would love to hear from @Dhaiwat10.

.husky/pre-commit
crondinini: @Nazeeh21 This was adding every file change made to the commit, even when I was selecting just some to be staged.

package.json
```diff
@@ -25,8 +29,8 @@
     "test:coverage": "jest --coverage --colors --maxWorkers=2",
     "hygen": "hygen",
     "component:new": "hygen component with-prompt",
-    "publish": "npx semantic-release --no-ci",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "publish": "npm publish"
```
with-heart, line 33: We should remove this line because it's redundant!

with-heart (approved):
Love the intent behind this PR and it's a good clean up!

I think we'd get more out of using `husky`+`lint-staged` so we can define our pre-commit tasks directly in `package.json`. Having them hidden in `.husky/pre-commit` isn't great imo.

This is great though!

package.json
crondinini: Yes, that's true! Will remove.

Re: Husky, I agree with using lint-staged with husky, that's how I've used husky in the past. Do you think it's worth doing it on this PR?

with-heart: Hmmmm. Actually now that you mention it, if you're down for it, wanna create a separate PR? It can serve as a shining example of what the transition from `husky` standalone to `husky`+`lint-staged` looks like.

Dhaiwat10:
@crondinini you can safely remove the `Example` component/folder. It's of no use to us now.

.nvmrc
```diff
@@ -1 +1 @@
-12
+16.13.0
```
Dhaiwat10, line 1: 👌

package.json
```diff
@@ -62,12 +59,10 @@
     "@types/react-dom": "^16.9.10",
     "babel-loader": "^8.2.1",
     "classnames": "^2.2.6",
-    "conventional-changelog-conventionalcommits": "^4.4.0",
-    "css-loader": "^5.0.1",
```
Dhaiwat10, line 66: Sure this won't break loading `.css` files?

Dhaiwat10 (approved):
Thank you!

package.json
crondinini: Sure! I can def do that :)

package.json
crondinini: Everything builds and storybook runs so for now yes. I never had to use a css loader in the past for loading CSS, I'm guessing it was always handled by the bundler, so maybe it could be done as part of the bundler config issue for example.

crondinini:
I've addressed all comments so I'll go ahead and merge this. :) 

Btw, I've enabled actions to run on pull requests so we can avoid getting to a bad merge stage.

PR #21, opened 2021-11-28
chore: use husky with lint staged

crondinini:
### Overview

This short PR introduces the usage of `lint-staged` in the husky pre-commit hook.

## TIL

TIL that Husky actually changed their approach to their Javascript configuration in v7.

Instead of it being done through `package.json` or an `.huskyrc.json`, it's now done with a bash script file within the `.husky` folder.

They explain the reasoning here:
https://blog.typicode.com/husky-git-hooks-javascript-config/

You can see the migration docs too:
https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v7

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn format
+yarn lint-staged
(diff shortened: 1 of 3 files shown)


Dhaiwat10 (approved):
Perfecto

PR #23, opened 2021-11-28
Update build config/scripts

etr2460:
This PR removes rollup, and instead builds with `tsc` only. It does the following:
- doesn't bundle the built assets together, and instead retains the current structure of importing from various directories/files within the repo
- Builds both for the commonjs and esm targets. I've seen this in other repos, and since i'm not 100% sure the best practice here, flexibility seemed ideal
- Updates a couple ts config/imports to make TSC work

A few things i'm not happy about, that might require future work in this PR or others:
- This doesn't handle non TS/TSX files. Will we have some of them in the future? I see we have an `assets` directory, so that may need to be managed
- The build command is rather long now (since i need to build for 2 targets and copy files after doing the build) and might be fragile? It's more for us to maintain vs. a library doing it for us
- I really don't know how to test this. Should i create my own private package from my fork and try to publish and use it? As a result, this might not work at all

New dist output:
image

to: @Dhaiwat10 @crondinini @with-heart


package.json
```diff
@@ -30,11 +30,12 @@
     "hygen": "hygen",
     "component:new": "hygen component with-prompt",
     "prepare": "husky install",
-    "pre-commit-hook": "yarn lint-staged"
+    "pre-commit-hook": "yarn lint-staged",
+    "tsc": "tsc --noEmit"
```
etr2460, line 34: this simply runs type checking without emitting built js assets

Dhaiwat10:
> This doesn't handle non TS/TSX files. Will we have some of them in the future?
We definitely might have to use some `.js` or images, etc as the project grows. 

> The build command is rather long now (since i need to build for 2 targets and copy files after doing the build) and might be fragile? It's more for us to maintain vs. a library doing it for us.
Not an expert on that department. @with-heart might have something to say.

> I really don't know how to test this.
In theory, as long as you are seeing a proper build output it's a good sign. I would suggest creating a copy of this and publishing it to npm as a dummy package or use something like `yarn link` to test locally.

etr2460:
>We definitely might have to use some .js or images, etc as the project grows.

I _think_ that js/jsx is allowed with the `allowJs` option, but i haven't tested. i'll need to try that here. As for images, i'm not 100% sure how that might work. My hope would be that we don't need any images other than SVGs, and in that case they could be embedded straight into the react component

etr2460:
closing in favor of preconstruct and #24

PR #25, opened 2021-11-29
Add test setup for monorepo and example test

etr2460:
I noticed we had no tests, so `yarn test` was failing (locally at least, in CI it was set to pass with no tests). This PR:
- adds an example test
- adds the jest-dom testing library helper functions (documentation here, but i've found them super useful: https://github.com/testing-library/jest-dom)
- removes the flag in CI to pass without tests (since now we have 1 test!)
- Moves TS and test infra out of `components` and into the root of the monorepo (so it can be shared by all packages)

to: @Dhaiwat10 @crondinini


Dhaiwat10:
This is something we absolutely need but I'm afraid you're gonna have to do a small re-factor in order to accomodate the new monorepo structure 😬

etr2460:
@Dhaiwat10 this is now rebased, and makes a couple additional changes to support a monorepo.

cc @crondinini to take a look at the monorepo parts (I think i'm understanding what stuff should be shared vs. split, but would be great to ensure this is following your mental model too)

package.json
```diff
@@ -8,7 +8,15 @@
     "pre-commit-hook": "yarn lint-staged",
     "format": "prettier --write",
     "build": "yarn workspace @web3-ui/components build",
-    "test": "yarn workspace @web3-ui/components test"
+    "test": "jest --maxWorkers=2",
+    "test:watch": "yarn test --watch",
+    "test:coverage": "jest --coverage --colors --maxWorkers=2"
```
with-heart, line 13: We should probably adjust our use of `--maxWorkers` flags here. `2` is oddly specific and imo doesn't maximize things in our favor: https://dev.to/vantanev/make-your-jest-tests-up-to-20-faster-by-changing-a-single-setting-i36

with-heart (approved):
Love it!

package.json
```diff
@@ -8,7 +8,15 @@
     "pre-commit-hook": "yarn lint-staged",
     "format": "prettier --write",
     "build": "yarn workspace @web3-ui/components build",
-    "test": "yarn workspace @web3-ui/components test"
+    "test": "jest",
```
crondinini, line 11: Yep, absolutely, that is what I was thinking about.

jest.config.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
+  preset: "ts-jest/presets/js-with-ts",
+  testMatch: [
+    "<rootDir>/packages/*/src/**/__tests__/**/*.{js,jsx,ts,tsx}",
+    "<rootDir>/packages/*/src/**/?(*.)(spec|test).{js,jsx,ts,tsx}",
+  ],
+  coveragePathIgnorePatterns: [
+    "/node_modules/",
+    "dist/",
```
crondinini, line 9: Would we need to ignore dist also from within the packages?

jest.config.js
```diff
@@ -0,0 +1,29 @@
...
+    "<rootDir>/packages/components/src/index.ts",
```
crondinini, line 10: Why is this needed if it doesn't match the spec?

crondinini (approved):
This is great! 🚀

jest.config.js
etr2460: this is for code coverage metrics. not sure why this file was specifically excluded before, but i kept it here to not break anything historical

jest.config.js
etr2460: i believe this will ignore any instance of `dist`, although i might be wrong. we're not using the coverage stats yet though, so it probably doesn't matter?

PR #26, opened 2021-11-29
chore: first step towards monorepo

crondinini:
### Overview

This PR is a first step for the monorepo. It's not the complete setup since it doesn't handle the publishing/building yet, and the hooks folder is just an example but it's possible to develop on the 2 packages separately as of the changes introduced. 

I chose to go forward with this despite it being incomplete so we can avoid major conflicts and we can free up people to work on things without the worry of having to solve those git conflicts. 

This is what the new folder structure looks like.

```
├── README.md
├── package.json
├── packages
│   ├── components
│   └── hooks
└── yarn.lock
```

### Changes

- Introduced a root `package.json`, which includes the `workspaces` reference + scripts necessary
- Moved lint staged to the root
- Introduced a fake hooks package and used it in the component package as an example
- Renamed packages to `@web3-ui/components` and `@web3-ui/hooks`
- Added the `@web3-ui/hooks` as a dependency of `@web3-ui/components`

### Missing

- Documentation around how to use it (will do it in my next PR)
   - You need to run `yarn` at the root level and it will link both packages
- Actually building the packages and publishing flow
- Test setup in the root folder (and maybe other root configs too)
- Proper setup of the hooks package (I didn't consider it to be in the scope of this PR)


package.json
```diff
@@ -1,82 +1,14 @@
...
-  "author": "Dhaiwat Pandya",
-  "homepage": "https://github.com/Developer-DAO/web3-ui/",
-  "bugs": {
-    "url": "https://github.com/Developer-DAO/web3-ui/issues"
-  },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/Developer-DAO/web3-ui/"
-  },
+  "private": "true",
```
etr2460, line 2: according to https://docs.npmjs.com/creating-a-package-json-file#required-name-and-version-fields, name and version are required fields in package.json files. maybe we should keep some stuff here still for whatever commands might be run from the root directory
```suggestion
  "name": "@web3-ui/root",
  "license": "MIT",  
  "version": "0.0.0",
  "private": "true",
  "engines": {
    "node": ">=16.0.0",
    "yarn": "^1.5"
  },
```

packages/components/package.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "@web3-ui/components",
+  "license": "MIT",
+  "version": "1.0.0",
```
etr2460, line 4: can we take advantage of the change in package name to revert these to a pre-release version number?
```suggestion
  "version": "0.0.1",
```

packages/components/package.json
```diff
@@ -0,0 +1,85 @@
...
+  "private": false,
+  "engines": {
+    "node": ">=16.0.0",
+    "yarn": "^1.5"
+  },
+  "description": "React Typescript Library Boilerplate",
```
etr2460, line 10:
```suggestion
  "description": "React components for web3, built with Chakra-UI",
```

packages/components/package.json
```diff
@@ -0,0 +1,85 @@
...
+  "author": "Dhaiwat Pandya",
+  "homepage": "https://github.com/Developer-DAO/web3-ui/",
+  "bugs": {
+    "url": "https://github.com/Developer-DAO/web3-ui/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Developer-DAO/web3-ui/"
+  },
+  "scripts": {
```
etr2460, line 20: maybe a future question, but do we want these scripts to live at the package level or the root level? I'd be a bit worried that the scripts for each package might diverge, and i'd rather us end up with something like `yarn test packages/components` at the root instead of `cd packages/components && yarn test` or `yarn workspace @web3-ui/components test`. food for thought

packages/components/package.json
```diff
@@ -0,0 +1,85 @@
...
+  "name": "@web3-ui/components",
+  "license": "MIT",
+  "version": "1.0.0",
+  "private": false,
+  "engines": {
+    "node": ">=16.0.0",
+    "yarn": "^1.5"
+  },
+  "description": "React Typescript Library Boilerplate",
+  "author": "Dhaiwat Pandya",
```
etr2460, line 11: should this be Developer DAO? Not sure what you think @Dhaiwat10 
```suggestion
  "author": "Developer DAO",
```

packages/hooks/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@web3-ui/hooks",
+  "version": "1.0.0",
```
etr2460, line 3:
```suggestion
  "version": "0.0.1",
```

packages/hooks/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "types": "index.d.ts",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
etr2460, line 12:
```suggestion
  "license": "MIT",
```

packages/hooks/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "name": "@web3-ui/hooks",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "types": "index.d.ts",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
```
etr2460, line 11:
```suggestion
  "author": "Developer DAO",
```

Dhaiwat10:
Thanks @crondinini, this is such a solid foundation for the project! Everything works perfectly. I am merging this in. If anyone wants to open a discussion regarding the monorepo setup or suggest improvements, please do so in this issue: #27.

packages/hooks/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@web3-ui/hooks",
+  "version": "1.0.0",
+  "description": "",
```
etr2460, line 4:
```suggestion
  "description": "React hooks for web3",
```

packages/hooks/package.json
```diff
@@ -0,0 +1,19 @@
+{
```
etr2460, line 1: looks like we're missing a decent amount of properties here that probably should be brought in from the components package.json too

etr2460 (commented):
a few comments, not sure how many should be addressed before merging though, makes sense to iterate and move fast

crondinini:
Hey @etr2460, I was fast asleep when the comments were made! I see you already addressed some of them, thank you for that. 

I think they all make sense, but this PR was more like an iterative step rather than a finished work. If we were working already with already published packages, I would have merged this into a feature branch for example.

package.json
crondinini: @etr2460 From reading the resource you shared, those are all required for a published package. Since the package is private it won't be published. Would you consider all of that still required?

This is not the final version for the package in any case, but just trying to understand.

packages/components/package.json
crondinini: Definitely and thanks for doing that. :)

packages/components/package.json
crondinini: I wanted to raise a similar discussion. Chakra has everything in the root folder, so the test and jest config are at the root and the packages only have the test file. So at the root we'd have `yarn test` and this would run all tests for example.

My comment in the PR description "Test setup in the root folder (and maybe other root configs too)", was referring to those sort of things so I'm 100% with you.

packages/hooks/package.json
crondinini: Yep, we are, as I mentioned in the PR description, this was a fake package, a POC of sorts. 
> "Introduced a fake hooks package and used it in the component package as an example"

Issue #27, opened 2021-11-29
Monorepo setup discussion


Dhaiwat10:
Please use this thread as the place to open a discussion about the monorepo setup (#26) or to suggest any improvements!

etr2460:
oops, looks like i was a minute or two late with PR comments. I might make a PR to add in the changes i listed here: https://github.com/Developer-DAO/web3-ui/pull/26#pullrequestreview-817348782

Dhaiwat10:
@etr2460 thank you! Please do that and I'll merge it in asap

etr2460:
opened https://github.com/Developer-DAO/web3-ui/pull/28!

Dhaiwat10:
#28 merged

crondinini:
Is there any further discussion needed? Seems like the improvements are in this other issue: 
https://github.com/Developer-DAO/web3-ui/issues/33

PR #29, opened 2021-11-29
Move LICENSE back to root and fix builds

etr2460:
Related to #27, we should have the LICENSE file in the root of the repo (so github picks it up), and we should make sure it gets pulled into the build output appropriately

to: @Dhaiwat10 @crondinini


packages/components/rollup.config.js
```diff
@@ -55,8 +55,8 @@ export default {
     }),
     copy({
       targets: [
-        { src: 'LICENSE', dest: 'dist' },
-        { src: 'README.md', dest: 'dist' },
+        { src: '../../LICENSE', dest: 'dist' },
+        { src: '../../README.md', dest: 'dist' },
```
Dhaiwat10, line 59: Don't we want the README to be separate for both the hooks and the components package? 🤔

etr2460: Maybe someday? but we only have one README file today :P So might as well use that shared one everywhere for now.

And this will probably all change with our new build/publishing system plans, this was mainly to keep the status quo from before the monorepo work

PR #31, opened 2021-11-29
feat: setup `hooks` package

Dhaiwat10:
This PR initalizes the `hooks` package and adds some starter code in the form of a `useWallet` hook.

## Known issues
- I tried using the `useWallet` hook inside the storybook for the `components` package. It did not work. For some reason, when I log `useWallet`'s contents all properties come out as `undefined`. It works fine in the storybook for the `hooks` package, though. Feel free to try both environments out yourself.

## Other changes
- Removed an unnecessary story from the `components` package
- Updated the story for `Address` in the`components` package
- Updated some scripts in the root `package.json`

## Notes
- I published `v0.1.0` of `@web3-ui/hooks` to test out if there's a problem with the bundling and stuff is just not being compiled.

cc @etr2460


Nazeeh21:
I tried this out.
While running `yarn`, it gave me the below error.
image

Then I upgraded my node to version 16, then the error disappeared.
Except that everything is working as mentioned in the PR.

Canopix (approved):

Canopix:
Great Work @Dhaiwat10 🚀

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@web3-ui/root",
   "license": "MIT",
   "version": "0.0.0",
-  "private": "true",
+  "private": "false",
```
with-heart, line 5: We want this to stay private, right? This should never be published because it's just a  way to coordinate the other packages.

packages/components/src/components/Address/Address.stories.tsx
```diff
@@ -2,13 +2,22 @@ import { storiesOf } from '@storybook/react';
 import React from 'react';
 import { Address } from '.';
 
-import { useWallet } from '@web3-ui/hooks';
+import { useWallet, Provider } from '@web3-ui/hooks';
+import { Button } from '@chakra-ui/react';
 
-storiesOf('Address', module).add('Default', () => <Address />);
+storiesOf('Address', module).add('Default', () => <Address value='This is an input' />);
```
etr2460, line 8: minor nit, but i like my storybook components to contain real looking content. So something like this might look better (and make it more clear what the component is for)
```suggestion
storiesOf('Address', module).add('Default', () => <Address value='testaddress.eth' />);
```

packages/hooks/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
etr2460, line 1: should prettier files be consolidated in the root of the monorepo too?

packages/hooks/LICENSE
```diff
@@ -0,0 +1,19 @@
+The MIT License
```
etr2460, line 1: this file is unneeded now that https://github.com/Developer-DAO/web3-ui/pull/29 is merged

packages/hooks/jest.config.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
```
etr2460, line 1: this is fine for now, but the approach in https://github.com/Developer-DAO/web3-ui/pull/25 would delete it in favor of the root jest config

packages/hooks/package.json
```diff
@@ -17,16 +17,64 @@
     "type": "git",
     "url": "git+https://github.com/Developer-DAO/web3-ui/"
   },
-  "main": "index.js",
   "scripts": {
-    "types": "index.d.ts",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "build": "yarn clean; rollup -c",
```
etr2460, line 21: lots of duplication between the components package here, but i mentioned all of that above

packages/hooks/package.json
```diff
@@ -17,16 +17,64 @@
...
+    "clean": "rimraf dist",
+    "format": "prettier --write \"src/**/*.{ts,tsx,json,js,jsx}\"",
+    "format:check": "prettier --list-different \"src/**/*.{ts,tsx,json,js,jsx}\"",
+    "test": "jest --maxWorkers=2",
+    "test:watch": "yarn test --watch",
+    "test:coverage": "jest --coverage --colors --maxWorkers=2",
+    "hygen": "hygen",
+    "component:new": "hygen component with-prompt",
```
etr2460, line 29: do we need this? it looks specific to components

packages/hooks/src/Provider.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export interface Web3ContextType {
+  connectWallet: (() => void) | undefined;
+  signer: JsonRpcSigner | null | undefined;
+  userAddress: string | null | undefined;
+  disconnectWallet: (() => void) | undefined;
+  network: string;
+  chainId: number | undefined;
+  connected: boolean;
+  provider: ethers.providers.Web3Provider | null | undefined;
```
etr2460, line 15: this is a styling preference, so no need to accept, but i find it easier to grok these when the shorthand for undefined is used, like this:
```suggestion
  connectWallet?: (() => void);
  signer?: JsonRpcSigner | null;
  userAddress?: string | null;
  disconnectWallet?: (() => void);
  network: string;
  chainId?: number;
  connected: boolean;
  provider?: ethers.providers.Web3Provider | null;
```

packages/hooks/package.json
```diff
@@ -17,16 +17,64 @@
...
+    "pre-commit-hook": "yarn lint-staged",
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook",
+    "dev": "yarn storybook"
```
with-heart, line 33: Similar to earlier comment but I think we shouldn't alias this to `dev`

packages/hooks/src/Provider.tsx
```diff
@@ -0,0 +1,94 @@
...
+ * @dev The global provider that handles and stores all the web3 connections. Wrap your entire App with this component.
+ * @param children Your app.
+ * @param network The network you want to connect to.
+ * @param infuraId Your Infura project ID. This is required if you want to support WalletConnect.
+ */
+export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId }) => {
+  const [signer, setSigner] = React.useState<null | JsonRpcSigner>(null);
+  const [provider, setProvider] = React.useState<null | ethers.providers.Web3Provider>(null);
+  const [userAddress, setUserAddress] = React.useState<null | string>(null);
+  const [web3Modal, setWeb3Modal] = React.useState<null | Web3Modal>();
```
etr2460, line 44: what's the difference between null and undefined here? if this is only ever undefined or `Web3Modal`, then lets remove the `null` type

packages/hooks/package.json
```diff
@@ -17,16 +17,64 @@
...
+  "types": "dist/index.d.ts",
   "dependencies": {
-    "react": "^17.0.2"
+    "@walletconnect/web3-provider": "^1.6.6",
+    "ethers": "^5.5.1",
+    "web3modal": "^1.9.4"
   },
   "peerDependencies": {
-    "react": ">=16.8.6"
+    "react": "*",
```
with-heart, line 44: At a minimum this needs to be `16.8.0` because that's when React added hooks support.

packages/hooks/src/hooks/useWallet.ts
```diff
@@ -0,0 +1,33 @@
...
+export function useWallet() {
+  const context = React.useContext(Web3Context);
+  const [ens, setEns] = React.useState<string>();
+
+  React.useEffect(() => {
+    if (context) {
+      const { userAddress, provider } = context;
+      if (userAddress && provider) {
+        provider.lookupAddress(userAddress).then((address) => {
+          setEns(address as string);
```
etr2460, line 16: do we need to handle an error case where the address isn't found?

packages/hooks/src/Provider.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const [chainId, setChainId] = React.useState<number>();
+  const [connected, setConnected] = React.useState<boolean>(false);
+
+  const connectWallet = React.useCallback(async () => {
+    console.log('connectWallet');
```
with-heart, line 49:
```suggestion
  const connectWallet = React.useCallback(() => {
```

packages/hooks/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
```
etr2460, line 1: if https://github.com/Developer-DAO/web3-ui/pull/25 goes in, then this file will probably get deleted too in favor of the root version

packages/hooks/src/Provider.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const signer = provider.getSigner();
+    setWeb3Modal(web3Modal);
+    setSigner(signer);
+    setProvider(provider);
+    setUserAddress(await signer.getAddress());
+    setChainId(await signer.getChainId());
+    setConnected(true);
+  }, [Web3Modal, web3Modal, WalletConnectProvider, network, infuraId, ethers]);
+
+  const disconnectWallet = React.useCallback(async () => {
```
with-heart, line 72:
```suggestion
  const disconnectWallet = React.useCallback(() => {
```

etr2460 (commented):
a few comments, nothing blocking, but mostly asking about how we want to handle the monorepo setup

packages/hooks/src/Provider.tsx
```diff
@@ -0,0 +1,94 @@
...
+      signer,
+      userAddress,
+      disconnectWallet,
+      connected,
+      provider,
+      network,
+      chainId,
+    }),
+    [connectWallet, signer, userAddress, web3Modal, connected, provider, network, chainId]
+  );
```
with-heart, line 91: `useMemo` probably isn't helping in the way you think it is.

Some resources that might explain why/provide alternatives:

- https://github.com/kentcdodds/old-kentcdodds.com/blob/319db97260078ea4c263e75166f05e2cea21ccd1/content/blog/how-to-optimize-your-context-value/index.md
- https://github.com/facebook/react/issues/15156#issuecomment-474590693
- https://kentcdodds.com/blog/how-to-optimize-your-context-value

I think we should just go with this for now and figure out a better pattern later, just wanted to throw that out there :)

packages/hooks/src/hooks/index.ts
```diff
@@ -0,0 +1 @@
+export * from './useWallet';
```
with-heart, line 1: One thing I've learned from experience is that it's better to be explicit with re-exports in `index` files:

```suggestion
export { useWallet } from './useWallet';
```

Anyone looking at this file can tell at a glance what is exported from this folder, whereas `export *` syntax requires them to read through each individual file to know what all is exported.

packages/hooks/src/Provider.tsx
```diff
@@ -0,0 +1,94 @@
+import { JsonRpcSigner } from '@ethersproject/providers/src.ts/json-rpc-provider';
+import React from 'react';
+import { ethers } from 'ethers';
+import WalletConnectProvider from '@walletconnect/web3-provider';
+import Web3Modal from 'web3modal';
+
+export interface Web3ContextType {
```
with-heart, line 7: > Web3ContextType

But it's not a type, it's an interface! 😆

packages/hooks/src/Provider.tsx
```diff
@@ -0,0 +1,94 @@
...
+  signer: JsonRpcSigner | null | undefined;
+  userAddress: string | null | undefined;
+  disconnectWallet: (() => void) | undefined;
+  network: string;
+  chainId: number | undefined;
+  connected: boolean;
+  provider: ethers.providers.Web3Provider | null | undefined;
+}
+
+export const Web3Context = React.createContext<Web3ContextType | undefined>(undefined);
```
with-heart, line 18: We'll probably want to use a few tricks for defining this context so we (or our users) don't use to check for its existence all over the place before we can use it. 

```ts
// without tricks 
const web3Context = useContext(Web3Context)
// gross
web3Context?.connectWallet()

// with tricks
const web3Context = useContext(Web3Context)
// yum
web3Context.connectWallet()
```

We'll get to it later!

packages/hooks/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './hooks';
+export * from './Provider';
```
with-heart, line 2: Similar to previous comment:

```suggestion
export { useWallet } from './hooks';
export {
  Provider,
  ProviderProps,
  Web3Context,
  Web3ContextType,
} from './Provider';
```

with-heart (approved):
A few suggestions but I think we could also just move forward with this and revisit things later. Whatever you wanna do 😁

package.json
```diff
@@ -15,8 +15,12 @@
     "prepare": "husky install",
     "pre-commit-hook": "yarn lint-staged",
     "format": "prettier --write",
-    "build": "yarn workspace @web3-ui/components build",
-    "test": "yarn workspace @web3-ui/components test"
+    "build:components": "yarn workspace @web3-ui/components build",
+    "build:hooks": "yarn workspace @web3-ui/hooks build",
```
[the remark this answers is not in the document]
Dhaiwat10: @etr2460 that's def a good idea but I suggest creating a separate issue/PR for that. Thoughts?

packages/hooks/src/Provider.tsx
Dhaiwat10: Yeah that def needs to be changed. Just laziness on my part. Wanted to ship this PR quick.

packages/hooks/package.json
Dhaiwat10: Probably not. I'll remove it

packages/hooks/src/hooks/useWallet.ts
Dhaiwat10: No. In that case, the ens will simply be null which is fine.

packages/hooks/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
...
+  printWidth: 100,
+  semi: true,
+  singleQuote: true,
+  trailingComma: 'es5',
+  jsxSingleQuote: true,
+};
```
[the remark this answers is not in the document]
Dhaiwat10: Great suggestion. I think this can be coupled with some other devops-y changes @etr2460 suggested in his reviews. We can batch these change together in a PR dedicated to that.

package.json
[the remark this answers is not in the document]
with-heart: Added this to the list in #33

package.json
```diff
@@ -15,8 +15,12 @@
...
+    "build": "yarn build:components && yarn build:hooks",
+    "test": "yarn workspace @web3-ui/components test",
+    "dev:components": "yarn workspace @web3-ui/components dev",
+    "dev:hooks": "yarn workspace @web3-ui/hooks dev"
```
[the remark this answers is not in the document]
with-heart: Created #32 and added to the list in #33

packages/hooks/.prettierrc.js
with-heart: Added to #33

packages/hooks/src/Provider.tsx
Dhaiwat10: Thanks for sharing these resources. Super helpful

packages/hooks/src/Provider.tsx
Dhaiwat10: classic

packages/hooks/src/Provider.tsx
Dhaiwat10: Yeah this is a massive pain point.

PR #40, opened 2021-11-30
Fix story for address with wallet

etr2460:
While verifying other changes, i noticed that the Address with wallet story wasn't working. that was because the provider was added after the `useWallet` hook. I fixed that, a type in the useWallet hook docstring, and ran `yarn install`

screenshots:
renders the modal
Screen Shot 2021-11-29 at 9 23 59 PM
shows the address on the testnet
Screen Shot 2021-11-29 at 9 24 03 PM
on mainnet shows my ENS address
Screen Shot 2021-11-29 at 9 24 19 PM

to: @Dhaiwat10


Dhaiwat10:
Oh..silly mistake. Good catch

PR #45, opened 2021-12-01
Feat/eth signin

RoguePenguins:
Just want to get this out so people can see it. I'm not great at react yet so there may be some bad design choices.


etr2460 (commented):
one thing to note here:

anything in a `*.stories.tsx` won't be included in the released package. Think of these files as building a sandbox for a package consumer to play around with to learn how the component works and what the different variations of the component are. therefore, all logic for the component needs to live outside the stories and test files and in the actual component file (in this case `Signin.tsx` instead)

RoguePenguins:
made new pull request with this repos branch

PR #48, opened 2021-12-01
Ethereum Signin

RoguePenguins:
This pull request adds a button to sign in with ethereum. It does have a bit of logic attached to it in the current form but I think I may split that into a hook to keep the ui library purely ui.

### DIFF
--- a/packages/components/src/components/Signin/Signin.stories.tsx
+++ b/packages/components/src/components/Signin/Signin.stories.tsx
@@ -1,0 +1,57 @@
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import { Signin } from '.';
+
+// builds EIP-4361 compliant message.
+function buildMessage(
+  domain: string,
+  address: string,
+  uri: string,
+  version: number,
+  chainId: number,
+  nonce: number,
+  issuedAt: string,
+  statement?: string,
+  requestId?: string,
+  exp?: Date,
+  notBefore?: Date,
+  resources?: string[]
+) {
+  let message = domain + ' wants you to sign in with your Ethereum account: \n' + address + '\n\n';
+
+  if (statement) message += statement + '\n\n';
+  message +=
+    'URI: ' +
+    uri +
+    '\nVersion: ' +
+    version +
+    '\nChain ID: ' +
+    chainId +
+    '\nNonce: ' +
+    nonce +
+    '\nIssued At: ' +
+    issuedAt;
+  if (exp) message += '\nExpiration Time: ' + exp;
+  if (notBefore) message += '\nNot Before' + notBefore;
+  if (requestId) message += '\nRequest Id: ' + requestId;
+  if (resources) message += '\nResources:' + resources.forEach((x) => (message += '\n- ' + x));
+
+  return message;
+}
+
+storiesOf('Sign-in', module).add('Default', () => <Signin data='Test' />);
+
+let now = new Date();
+const message = buildMessage(
+  'example.xyz',
+  '0x000000000000000',
+  'https://sign-in',
+  1,
+  1,
+  1234,
+  now.toISOString(),
+  'This is a test sign-in',
+  '1'
+);
+
+storiesOf('Sign-in', module).add('Standard Message', () => <Signin data={message} />);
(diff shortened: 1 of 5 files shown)


RoguePenguins:
Also does anyone understand the error in the build? It works how I expect in the storybook but that error has me stumped.

RoguePenguins:
> Also does anyone understand the error in the build? It works how I expect in the storybook but that error has me stumped.

got it, just had to move dependencies to the right place

packages/components/src/components/Signin/Signin.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+  chainId: number,
+  nonce: number,
+  issuedAt: string,
+  statement?: string,
+  requestId?: string,
+  exp?: Date,
+  notBefore?: Date,
+  resources?: string[]
+) {
+  let message = domain + ' wants you to sign in with your Ethereum account: \n' + address + '\n\n';
```
brandon-kyle-bailey, line 20: Could we use a string literal template her instead ?

Dhaiwat10:
Thanks for the PR  @RoguePenguins! This is a great idea but there are some things that I want to discuss:

Like you said, this component contains logic which is not exactly what we want in our `components` package. I think this _does_ perfectly fit in to the use case of `web3-ui` - the third package that I was planning which would be a opinionated combination of our `components` and `hooks`. I might initialize that package soon and will mention you here so that we can move this to that package. :)

packages/components/src/components/Signin/Signin.stories.tsx
RoguePenguins: ahh yes we could. I did a lot of nonsense in school and to get better at practical coding lol. I appreciate the advice!

crondinini:
@Dhaiwat10 Instead of waiting until we have this other combined library, I'd say we split this out into a hook as @RoguePenguins mentioned the PR description.

Dhaiwat10:
@crondinini that's a great suggestion. Let's do that

packages/components/src/components/Signin/Signin.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Button } from '@chakra-ui/react';
+export interface SigninProps {
+  data: string;
+}
+
+export const Signin: React.FC<SigninProps> = (props) => {
+  const { connected, connectWallet } = useWallet();
+
+  async function handleSignIn() {
+    if (!connected) connectWallet;
```
crondinini, line 13: Isn't `connectWallet` a function?

packages/components/src/components/Signin/Signin.tsx
```diff
@@ -0,0 +1,26 @@
+import { ethers } from 'ethers';
+import React from 'react';
+import { useWallet, Provider } from '@web3-ui/hooks';
+import { Button } from '@chakra-ui/react';
+export interface SigninProps {
+  data: string;
```
crondinini, line 6: Seems like `data` can hold different types, not just a string, could we use a generic to allow folks to pass what is the expected data?

packages/components/src/components/Signin/Signin.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+  uri: string,
+  version: number,
+  chainId: number,
+  nonce: number,
+  issuedAt: string,
+  statement?: string,
+  requestId?: string,
+  exp?: Date,
+  notBefore?: Date,
+  resources?: string[]
```
crondinini, line 18: Not major since this is only in a story, but in general, when there are so many params, I'd suggest making this an object for ease of calling the function.

crondinini (commented):
Thanks for the PR! Added a couple of comments. Some of it still applies in the case of extracting this to the hook.

packages/components/src/components/Signin/Signin.stories.tsx
RoguePenguins: thank you for the feedback! It will definitely help me rebuild some of this better in a hook

packages/components/src/components/Signin/Signin.tsx
RoguePenguins: so data is the message the user signs in their wallet but it should be named more clearly.

Dhaiwat10:
@RoguePenguins now that we have the `core` package in place, would you be interested in adding sign in with ethereum to that package?

Dhaiwat10:
Zzzz

PR #60, opened 2021-12-03
Consolidate storybook and update docs

### LINKED ISSUE #32: Move to a single root Storybook instance
We've currently set up each of our packages with its own Storybook config but I'd like to suggest that we move to a single root instance as multiple Storybook instances are unnecessary complexity for no real gain. 

A single instance is useful for local dev (run once to explore all packages as we make changes) and can be published as a single point for users looking to explore the visual output of all of our packages. 

The single root instance can be configured in an almost identical way to the current configs, except `stories` in `.storybook/main.js` should point to `../packages/**/src/*.stories.@(js|jsx|ts|tsx|mdx)` (or maybe just `../packages/**/*.stories.@(js|jsx|ts|tsx|mdx)` so we can organize the stories however we like, even outside of `src` folders if we want).

etr2460:
See title, Closes #32

to: @crondinini @Dhaiwat10 @with-heart


Dhaiwat10 (approved):
Awesome man LGTM

PR #46, opened 2021-12-01
Add NFT Gallery and NFT components

etr2460:
This PR adds 2 new components to web3-ui: `NFTGallery` and `NFT`

NFTGallery calls the opensea API to fetch all the assets owned by a given wallet. It then uses the NFT component to render these in a grid.

A couple limitations:
- Requires on the opensea API (sooo web2) because there isn't a decent way of getting all tokens owned by a wallet
- Only works with image NFTs
- Shows minimal metadata about the NFT

Obviously theres a bunch of future work that can be done here, but this was me getting to learn how to use Chakra-UI mainly. Some future enhancements:
- Add a wrapper around NFT (and make the current implementation private) that uses a tokenId and contract address to fetch metadata about it
- Support other NFT types (audio, video, etc.)
- add more metadata about the nfts

## Test Plan
Added new stories and unit tests. This included the introduction of `msw` for mocking API responses ([thanks Kent C. Dodds](https://kentcdodds.com/blog/stop-mocking-fetch)).

Screen Shot 2021-11-30 at 11 38 23 PM
Screen Shot 2021-11-30 at 11 38 29 PM

to: @Dhaiwat10 @crondinini @with-heart

PR #62, opened 2021-12-03
NFT component improvements

### LINKED ISSUE #61: NFT component improvements
- [ ] ENS support
- [ ] GIF support
- [ ] Video support
- [ ] Audio support

Dhaiwat10:
- [x] ENS support
- [x] GIF support
- [x] Video support
- [x] Audio support
- [x] Fix the tests
- [x] Fetch data more efficiently. At the moment, both the Gallery and the NFT components are fetching the same data twice.

Also refactored the `<NFT />` component to expect `contractAddress` and `tokenId` instead of accepting `imageUrl` etc. 

Closes #61


packages/components/src/components/NFT/NFT.tsx
```diff
@@ -15,24 +22,66 @@ export interface NFTProps {
...
  */
-export const NFT = ({
-  tokenId,
-  name,
-  imageUrl,
-  assetContractName,
-  assetContractSymbol,
-}: NFTProps) => {
-  const displayName = name || tokenId;
+export const NFT = ({ contractAddress, tokenId }: NFTProps) => {
```
etr2460, line 25: do you think we could take the approach i mentioned here for the NFT component:https://github.com/Developer-DAO/web3-ui/pull/46#issue-1068068696
>Add a wrapper around NFT (and make the current implementation private) that uses a tokenId and contract address to fetch metadata about it

We should keep the existing component around and use it for the `NFTGallery` component so we don't need to refetch the NFT data that already comes from the previous API

packages/components/src/components/NFT/NFT.tsx
```diff
@@ -15,24 +22,66 @@ export interface NFTProps {
  * TODO: Add a component that fetchs the NFT data from the blockchain and uses this component to
```
etr2460, line 22: let's cleanup this TODO (now that it's done)

packages/components/src/components/NFT/NFT.tsx
```diff
@@ -1,12 +1,19 @@
...
   tokenId: string;
-  name: string | null;
-  imageUrl: string;
-  assetContractName: string;
+}
+
+interface NFTData {
+  tokenId: string;
+  imageUrl?: string;
+  name: string;
```
etr2460, line 13: not every NFT has a name. that's why this was nullable and i had `const displayName = name || tokenId;` below

packages/components/src/components/NFT/NFT.tsx
```diff
@@ -15,24 +22,66 @@ export interface NFTProps {
...
+  useEffect(() => {
+    _isMounted.current = true;
+    fetchNFTData();
+    return () => {
+      _isMounted.current = false;
+    };
+  }, [contractAddress, tokenId]);
+
+  if (!nftData) {
+    return null;
```
etr2460, line 54: maybe let's render a loader with a box around it the size of the NFT card by default, so that we don't have the cards pop in and move the layout around?

packages/components/src/components/NFT/NFT.tsx
```diff
@@ -15,24 +22,66 @@ export interface NFTProps {
...
-  assetContractSymbol,
-}: NFTProps) => {
-  const displayName = name || tokenId;
+export const NFT = ({ contractAddress, tokenId }: NFTProps) => {
+  const _isMounted = useRef(true);
+  const [nftData, setNftData] = React.useState<NFTData>();
+
+  const fetchNFTData = useCallback(async () => {
+    const response = await (
+      await fetch(`https://api.opensea.io/api/v1/asset/${contractAddress}/${tokenId}/`)
```
etr2460, line 31: don't forget error handling!

Also, we should decide as a group if we wanna use `.then` or `async/awaits` for fetching. i'm fine either way, but let's be consistent

utils/server-handlers.ts
```diff
@@ -1,16 +1,19 @@
 // File with API handlers shared across all tests.
 // See https://kentcdodds.com/blog/stop-mocking-fetch for more details
 
-import { rest } from "msw";
-import { MOCK_OPENSEA_ASSETS_RESPONSE } from "./mocks";
+import { rest } from 'msw';
```
etr2460, line 4: weird that prettier didn't prettify this originally. must be a config issue

etr2460 (commented):
a few comments, looking awesome!

packages/components/src/components/NFT/NFT.tsx
Dhaiwat10: @etr2460 missed that. That sounds fair

packages/components/src/components/NFT/NFT.tsx
Dhaiwat10: Gotcha

packages/components/src/components/NFT/NFT.tsx
Dhaiwat10: Yep, good idea

packages/components/src/components/NFT/NFT.tsx
Dhaiwat10: Yeah, will add error handling.

I prefer `async/await`. Wbu? And @crondinini?

etr2460: me too, let's go with `async/await`. i was probably a bit lazy on my other PR to not factor that out. Maybe it's worth moving all the api requests into their own file too, that way any component can import them.

crondinini: I'm more of an async/await kinda of person. But I'm fine either way as long as we have consistency.

In this case though, I don't think we need the 2 awaits. As far as I know `.json()` does not return a promise. I also think we don't need `useCallback` at this point. 

Could do:
```js
const response = (
      await fetch(`https://api.opensea.io/api/v1/asset/${contractAddress}/${tokenId}/`)
    ).json();
```
or 

```js
  const fetchNFTData = async () => {
    const fetchResponse = await fetch(`https://api.opensea.io/api/v1/asset/${contractAddress}/${tokenId}/`);

    if (_isMounted.current) {
      const response = fetchResponse.json();
      setNftData({
        tokenId: response.token_id,
        imageUrl: response.image_url,
        name: response.name,
        assetContractName: response.asset_contract.name,
        assetContractSymbol: response.asset_contract.symbol,
        animationUrl: response.animation_url,
      });
    }
  };
```

etr2460: you should need 2 awaits. `json()` is async, because parsing the response can be expensive for extremely large responses

PR #68, opened 2021-12-03
Feat: Add custom Web3Modal Wallet Providers

### LINKED ISSUE #75: Custom Web3Modal providers
In the **useWallet** hook, currently, we are providing support for Metamask and WalletConnect providers.
I have created PR #68 to add support for more providers.

This issue is created to discuss whether to provide support to pass any (custom) Web3Modal providers the user pleases. This is the continuation of the discussion from PR #68.

Nazeeh21:
Closes #75

## Description
- I've added Coinbase and Ledger wallet Providers in useWallet hook.
- While connecting using *WalletConnect*, previously it was throwing an error that RPC not specified. To fix that, I've added RPC to the *WalletConnet* provider options.

## 📝 Additional Information

We can accept props from the user that which wallet provider they want to use. Also, we can add support to add custom wallet providers.

## Screenshot
image

### DIFF
--- a/packages/hooks/src/Provider.tsx
+++ b/packages/hooks/src/Provider.tsx
@@ -50,6 +50,7 @@ export interface ProviderProps {
  * @param children Your app.
  * @param network The network you want to connect to.
  * @param infuraId Your Infura project ID. This is required if you want to support WalletConnect.
+ * @param extraWalletProviders An array of extra Wallet Providers you want to support.
  */
 export const Provider: React.FC<ProviderProps> = ({
   children,
--- a/packages/hooks/src/Provider.tsx
+++ b/packages/hooks/src/Provider.tsx
@@ -1,10 +1,10 @@
 import { JsonRpcSigner } from '@ethersproject/providers/src.ts/json-rpc-provider';
-import React from 'react';
-import { ethers } from 'ethers';
-import WalletLink from 'walletlink';
 import { ledgerProviderOptions } from '@rsksmart/rlogin-ledger-provider';
 import WalletConnectProvider from '@walletconnect/web3-provider';
-import Web3Modal from 'web3modal';
+import { ethers } from 'ethers';
+import React from 'react';
+import WalletLink from 'walletlink';
+import Web3Modal, { IProviderOptions } from 'web3modal';
 
 export interface Web3ContextType {
   connectWallet?: () => void;
@@ -31,6 +31,8 @@ export interface ProviderProps {
    * @type string
    */
   infuraId?: string;
+
+  extraWalletProviders: [IProviderOptions];
 }
 
 /**
@@ -39,7 +41,12 @@ export interface ProviderProps {
  * @param network The network you want to connect to.
  * @param infuraId Your Infura project ID. This is required if you want to support WalletConnect.
  */
-export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId }) => {
+export const Provider: React.FC<ProviderProps> = ({
+  children,
+  network,
+  infuraId,
+  extraWalletProviders,
+}) => {
   const [signer, setSigner] = React.useState<null | JsonRpcSigner>();
   const [provider, setProvider] = React.useState<ethers.providers.Web3Provider | null>();
   const [userAddress, setUserAddress] = React.useState<null | string>();
@@ -90,6 +97,7 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId 
         'custom-ledger': {
           ...ledgerProviderOptions,
         },
+        ...extraWalletProviders[0],
       },
     });
     setWeb3Modal(web3Modal);
--- a/packages/hooks/src/Provider.tsx
+++ b/packages/hooks/src/Provider.tsx
@@ -1,6 +1,8 @@
 import { JsonRpcSigner } from '@ethersproject/providers/src.ts/json-rpc-provider';
 import React from 'react';
 import { ethers } from 'ethers';
+import WalletLink from 'walletlink';
+import { ledgerProviderOptions } from '@rsksmart/rlogin-ledger-provider';
 import WalletConnectProvider from '@walletconnect/web3-provider';
 import Web3Modal from 'web3modal';
 
@@ -48,13 +50,48 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId 
   const [connection, setConnection] = React.useState<any>();
 
   const connectWallet = React.useCallback(async () => {
+    // Coinbase walletLink init
+    const walletLink = new WalletLink({
+      appName: 'coinbase',
+    });
+
+    // WalletLink provider
+    const walletLinkProvider = walletLink.makeWeb3Provider(
+      'https://eth-mainnet.alchemyapi.io/v2/qCdzfF9UqXcJYIle-Ff-BN0MII8LjLQs',
+      1
+    );
+
     const web3Modal = new Web3Modal({
       providerOptions: {
         walletconnect: {
           package: WalletConnectProvider,
           options: {
+            bridge: 'https://polygon.bridge.walletconnect.org',
             infuraId,
+            rpc: {
+              1: 'https://eth-mainnet.alchemyapi.io/v2/qCdzfF9UqXcJYIle-Ff-BN0MII8LjLQs', // mainnet // For more WalletConnect providers: https://docs.walletconnect.org/quick-start/dapps/web3-provider#required
+              42: infuraId,
+              100: 'https://dai.poa.network', // xDai
+            },
           },
+        },
+        'custom-walletlink': {
+          display: {
+            logo: 'https://play-lh.googleusercontent.com/PjoJoG27miSglVBXoXrxBSLveV6e3EeBPpNY55aiUUBM9Q1RCETKCOqdOkX2ZydqVf0',
+            name: 'Coinbase',
+            description: 'Connect to Coinbase Wallet (not Coinbase App)',
+          },
+          package: walletLinkProvider,
+          // @ts-ignore
+          connector: async (provider, options) => {
+            await provider.enable();
+            return provider;
+          },
+        },
+        'custom-ledger': {
+          ...ledgerProviderOptions,
+          // rpc: { 31: 'https://public-node.testnet.rsk.co' },
+          // chainId: 31,
         },
       },
     });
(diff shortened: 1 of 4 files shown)


packages/hooks/src/Provider.tsx
```diff
@@ -46,15 +48,50 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId
...
+          // @ts-ignore
+          connector: async (provider, options) => {
+            await provider.enable();
+            return provider;
+          },
+        },
+        'custom-ledger': {
+          ...ledgerProviderOptions,
+          // rpc: { 31: 'https://public-node.testnet.rsk.co' },
+          // chainId: 31,
```
etr2460, line 93: should we remove these comments?

packages/hooks/src/Provider.tsx
```diff
@@ -46,15 +48,50 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId
...
 
   const connectWallet = React.useCallback(async () => {
+    // Coinbase walletLink init
+    const walletLink = new WalletLink({
+      appName: 'coinbase',
+    });
+
+    // WalletLink provider
+    const walletLinkProvider = walletLink.makeWeb3Provider(
+      'https://eth-mainnet.alchemyapi.io/v2/qCdzfF9UqXcJYIle-Ff-BN0MII8LjLQs',
```
etr2460, line 58: i see this url used in 2 places, let's pull it out into a constant?

packages/hooks/src/Provider.tsx
```diff
@@ -46,15 +48,50 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId
...
+            rpc: {
+              1: 'https://eth-mainnet.alchemyapi.io/v2/qCdzfF9UqXcJYIle-Ff-BN0MII8LjLQs', // mainnet // For more WalletConnect providers: https://docs.walletconnect.org/quick-start/dapps/web3-provider#required
+              42: infuraId,
+              100: 'https://dai.poa.network', // xDai
+            },
           },
         },
+        'custom-walletlink': {
+          display: {
+            logo: 'https://play-lh.googleusercontent.com/PjoJoG27miSglVBXoXrxBSLveV6e3EeBPpNY55aiUUBM9Q1RCETKCOqdOkX2ZydqVf0',
```
etr2460, line 79: let's grab a coinbase svg from somewhere and move it into the repo, that way we don't need to rely on a random google url

packages/hooks/src/Provider.tsx
```diff
@@ -46,15 +48,50 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId
...
+            name: 'Coinbase',
+            description: 'Connect to Coinbase Wallet (not Coinbase App)',
+          },
+          package: walletLinkProvider,
+          // @ts-ignore
```
etr2460, line 84: instead of `@ts-ignore`, can we use `@ts-expect-error` and include a comment about why this is needed?

packages/hooks/src/Provider.tsx
```diff
@@ -46,15 +48,50 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId
...
     const web3Modal = new Web3Modal({
       network,
       providerOptions: {
         walletconnect: {
           package: WalletConnectProvider,
           options: {
+            bridge: 'https://polygon.bridge.walletconnect.org',
             infuraId,
+            rpc: {
+              1: 'https://eth-mainnet.alchemyapi.io/v2/qCdzfF9UqXcJYIle-Ff-BN0MII8LjLQs', // mainnet // For more WalletConnect providers: https://docs.walletconnect.org/quick-start/dapps/web3-provider#required
```
etr2460, line 71: should we add more than just mainnet here, for the other networks that the user can pass in?

etr2460 (commented):
a few comments, but i'll leave stamping up to @Dhaiwat10 since he's got the most context on the hooks/provider side of stuff

Dhaiwat10:
Me and @Nazeeh21 had a discussion a few days ago about letting the user pass in these providers as they please. That way we will decrease the dependenices of this package, decrease the bundle size and not restrict the users to the providers we have chosen and hardcoded into our source code. @etr2460 what are your thoughts on that?

packages/hooks/src/Provider.tsx
narbs91: Did you all really want to expose that endpoint out in the wild?  Theoretically anyone can steal that url and use it for themselves.  I would suggest perhaps hiding it in an env variable is possible or let the consumer pass in there own rpc link

etr2460:
>Me and @Nazeeh21 had a discussion a few days ago about letting the user pass in these providers as they please. That way we will decrease the dependenices of this package, decrease the bundle size and not restrict the users to the providers we have chosen and hardcoded into our source code. @etr2460 what are your thoughts on that?

As long as the Providers all share an API, then that's probably fine? I assume we'd need to have at least one for testing in storybook still, but other than that, it seems reasonable. You both probably know more about this topic than I do too, so i'll defer to your judgement here

packages/hooks/src/Provider.tsx
Nazeeh21: Yes, I'll remove that

packages/hooks/src/Provider.tsx
Nazeeh21: Yes, I agree, we need to accept this as a prop from the user. I'm using it here for development purposes. 
It's not even mine lol, I got that from one from the repo I used as a ref.

packages/hooks/src/Provider.tsx
Nazeeh21: Sure, I'll do it that way.

packages/hooks/src/Provider.tsx
Nazeeh21: Yes, that could be done.

packages/hooks/src/Provider.tsx
Nazeeh21: I do think it could be done. I'll probably have to do some more research and find a way to do it.

Dhaiwat10:
> As long as the Providers all share an API
I think all of them share the same API in web3modal, yeah. Should be easy to integrate that. What are your thoughts @Nazeeh21?

packages/hooks/src/Provider.tsx
narbs91: Cool, cool yeah these rpc links seem to magically appear in all the projects lol

christiananese: I think it makes if we could define and agree on some defaults. It makes sense to define URLs and constants in a common file if they can be exposed. In my opinion it makes the code more beginner friendly and easier to read. 

The second argument for the `makeWeb3Provider` is the chainId, I think a it's hard for a new user to understand waht it is and to imply that he knows that the chainID 1 is for ETH mainnet.

Nazeeh21:
Yes, definitely. Let me share a quick roadmap here.
We'll be allowing the following providers from the user as props and docs for the same.
- [WalletConnect](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/walletconnect.md)
- [Fortmatic](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/fortmatic.md)
- [Torus](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/torus.md)
- [Portis](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/portis.md)
- [Authereum](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/authereum.md)
- [Frame](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/frame.md)
- [Bitski](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/bitski.md)
- [Arkane](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/arkane.md)
- [DCent](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/dcent.md)
- [BurnerConnect](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/burnerconnect.md)
- [MEWConnect](https://github.com/web3modal/web3modal/blob/HEAD/docs/providers/mewconnect.md)

Nazeeh21:
> We'll be allowing the following providers from the user as props and docs for the same.

We'll be having the same integration docs as the Web3modal because we'll be passing these providers directly to the Web3modal using their same syntax.

Dhaiwat10:
@Nazeeh21 how did you come up with this list?

Dhaiwat10:
And what do you mean by docs? The API for  adding any of these providers is the same, right? Why do we need different docs for all of them?

Nazeeh21:
> @Nazeeh21 how did you come up with this list?
I found this list on the [Web3Modal docs](https://www.npmjs.com/package/web3modal).

> And what do you mean by docs? The API for adding any of these providers is the same, right? Why do we need different docs for all of them?
Yes, for adding any of these providers, the API is the same.

Dhaiwat10:
@Nazeeh21 great. Can you please create an issue regarding being able to pass any Web3Modal providers the user pleases? Let's take the discussion there since I have some pointers to share.

Nazeeh21:
Sure 😊

Dhaiwat10:
@Nazeeh21 can you get rid of the comments and replace that infura ID with the `infuraId` we are accepting to the `<Provider />`?

Nazeeh21:
Yes, I'm working on removing the comments and I'll replace that infuraId with the `infuraId` from the `<Provider />`

Nazeeh21:
@Dhaiwat10 I've implemented the basic functionality of accepting `extraWalletProviders` from the users. Please review it and provide your suggestions. : )

packages/hooks/src/Provider.tsx
```diff
@@ -48,15 +68,49 @@ export const Provider: React.FC<ProviderProps> = ({ children, network, infuraId
...
+
+    const defaulProviderOptions = {
+      walletconnect: {
+        package: WalletConnectProvider,
+        options: {
+          bridge: 'https://polygon.bridge.walletconnect.org',
+          infuraId,
+          rpc: {
+            1: `https://eth-mainnet.alchemyapi.io/v2/${infuraId}`, // mainnet // For more WalletConnect providers: https://docs.walletconnect.org/quick-start/dapps/web3-provider#required
+            42: infuraId,
```
Dhaiwat10, line 90: Wouldn't we need to append a link to the `infuraId` here? Just like we did for mainnet?

Nazeeh21: Thanks for pointing it out. It just went out of my mind. Adding that.

PR #69, opened 2021-12-04
Feat/use token balance

AlexNi245:
## Description
A simple hook that fetches the user's balance of a certain ERC20 token. 

## Aditional Information
I created 3 testcases in the storybook section
1. Success -> Owner and Contract Id's are valid
2. Error -> undefined contract address
3. Error -> undefined owner address

I want to write unit test in the first place but still not sure how to mock the web3context within a unit test :( so I decided to used the storybook instead


packages/hooks/src/hooks/useTokenBalance/ERC20.json
```diff
@@ -0,0 +1,222 @@
+[
```
etr2460, line 1: Did you write this, or was it generated/sourced from elsewhere? I wonder if we can import this from a library instead of needing to define the schema for an ERC20 contract

packages/hooks/src/hooks/useTokenBalance/index.ts
```diff
@@ -0,0 +1 @@
+export { useTokenBalance } from './useTokenBalance';
```
etr2460, line 1: i think you also need to export this from the root hooks directory here: https://github.com/Developer-DAO/web3-ui/blob/master/packages/hooks/src/index.ts

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,45 @@
+import React, { useContext } from 'react';
+import { ethers } from 'ethers';
+import { Web3Context } from '../../Provider';
+import ERC20 from './ERC20.json';
+
+export const useTokenBalance = (address: string, owner: string) => {
```
etr2460, line 6: not sure what naming conventions are common here, but maybe `contractAddress`? I could also see us wanting `ownerAddress` instead of `owner`, unless we also accept ENS domains for the owner string

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,45 @@
...
+  const [isLoading, setIsLoading] = React.useState(true);
+  const [balance, setBalance] = React.useState<null | string>(null);
```
etr2460, line 8: This should be fine, and will default `balance` to `undefined`
```suggestion
  const [balance, setBalance] = React.useState<string>();
```

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,45 @@
...
+  const [hasError, setHasError] = React.useState(false);
```
etr2460, line 9: maybe it's better to make this `errorMessage` and set the type to `string`. Then, depending on the error, we can pass a different message up

packages/hooks/src/stories/UseTokenBalance.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+storiesOf('useTokenBalance', module).add('Default', () => (
+  <Provider network='mainnet'>
+    <Default />
+  </Provider>
+));
+
+const BalanceOfDai = () => {
+  const owner = '0x503828976D22510aad0201ac7EC88293211D23Da';
+  const address = '0x6B175474E89094C44Da98b954EedeAC495271d0F';
```
etr2460, line 30: can we pull these strings into constants at the top of the file (since they're reused) and name them appropriately? I assume that `address` is `DAI_CONTRACT_ADDRESS`?

packages/hooks/src/stories/UseTokenBalance.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <UnknownContract />
+      </>
+    );
+  }
+
+  return <button onClick={connectWallet}>Connect Wallet</button>;
+};
+
+storiesOf('useTokenBalance', module).add('Default', () => (
+  <Provider network='mainnet'>
```
etr2460, line 23: should we use a non mainnet for stories? I think we used rinkeby for other stories

etr2460 (commented):
a few comments. i think it's probably ok to not have unit tests for now since you added the storybook. but i'll need to look into other possible tests at some point. there's got to be a good way to mock the provider

packages/hooks/src/hooks/useTokenBalance/ERC20.json
AlexNi245: i copied it from the following  [sitw](https://ethereumdev.io/abi-for-erc20-contract-on-ethereum/) but I've you know a library that provides the interface we should replace the .json file

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
AlexNi245: I think the obvious the function names the better. So I also think that  ownerAddress and contractAdrees are the better options for this. 
But haven't tested the hook with an ens domain yet so I don't know how it behaves in this case.

etr2460: it probably doesn't support ENS unless you add it. I think we've implemented ENS support in other hooks/components already, so maybe it makes sense to pull that code out into a util/hook and reuse it here too. ENS support isn't blocking though

packages/hooks/src/hooks/useTokenBalance/ERC20.json
etr2460: I looked around, but couldn't find a legit looking library for this anywhere (i found one with only 1 star, but that seems risky to rely on).

Maybe let's rename the file to `ERC20ABI.json` at least, for a bit more clarity? I assume it'll never need to change (i'm still a bit new to web3, so don't have a full understanding of this yet)

AlexNi245: Yes the ERC20 interface will not change over time. Will rename the file to achieve more clearity

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
AlexNi245: i think util for ens would make sense here. I can have a look at how to do this best after this one is merged.

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @param contractAddress -> the address of the ERC20 token contract
+ * @param ownerAddress -> the address of the owner which balance should be fetched
+
+ */
+export const useTokenBalance = (contractAddress: string, ownerAddress: string) => {
+  const [isLoading, setIsLoading] = React.useState(true);
+  const [balance, setBalance] = React.useState<string>();
+  const [errorMsg, setErrorMsg] = React.useState(false);
```
Dhaiwat10, line 14:
```suggestion
  const [error, setError] = React.useState(false);
```

I think `error` is a better name

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Web3Context } from '../../Provider';
+import ERC20 from './ERC20ABI.json';
+
+/**
+ * @param contractAddress -> the address of the ERC20 token contract
+ * @param ownerAddress -> the address of the owner which balance should be fetched
+
+ */
+export const useTokenBalance = (contractAddress: string, ownerAddress: string) => {
+  const [isLoading, setIsLoading] = React.useState(true);
```
Dhaiwat10, line 12:
```suggestion
  const [loading, setLoading] = React.useState(true);
```

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,52 @@
...
+
+      setBalance(displayBalance);
+      setIsLoading(false);
+    } catch (e) {
+      setErrorMsg(e);
+      setIsLoading(false);
+    }
+  };
+
+  return [isLoading, balance, errorMsg];
```
Dhaiwat10, line 51:
```suggestion
  return { balance, loading, error };
```

Let's follow the convention used in the other hook!

Dhaiwat10 (changes requested):
Thanks for this PR @AlexNi245! I have requested a few changes. Can you please look into them?

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
AlexNi245: Is it okay for you @etr2460 ? If so I will rename it to "error"

Dhaiwat10:
@AlexNi245 the storybook setup needs some changes. I'm gonna be working on it and pushing commits to this PR. So if you're working on this PR rn, please keep that on hold!

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,51 @@
...
+      setLoading(true);
+      try {
+        getBalance();
+      } catch (error) {
+        console.error(error);
+        setError(true);
+      }
+      setLoading(false);
+    }
+  }, [tokenAddress, accountAddress]);
```
Dhaiwat10, line 41: @AlexNi245 this is one thing that I want you to look at. I just fixed this but you didn't have a dependency array attached to your `useEffect` which was causing it to run in an infinite loop.

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  return {
+    balance: balance?.toString(), // The balance in wei
+    loading,
+    error,
+    decimals,
+    formattedBalance: balance && ethers.utils.formatUnits(balance, decimals), // The balance in ethers eg. 0.01 ETH, 20 GTC, etc.
+    balanceInBigNumber: balance,
```
Dhaiwat10, line 49: I added these two additional helper variables.

tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "pretty": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es5"
+    "target": "es5",
+    "resolveJsonModule": true
```
Dhaiwat10, line 21: Added this to support `.json` imports

packages/hooks/src/hooks/useTokenBalance/useTokenBalance.ts
etr2460: My thought was that this shouldn't be a boolean at all, but instead be an errorMessage and a string type, so that we can pass a better error message back to the caller:
>maybe it's better to make this errorMessage and set the type to string. Then, depending on the error, we can pass a different message up


I can iterate on this in a future PR though

Dhaiwat10:
@allcontributors please add @AlexNi245 for code

PR #74, opened 2021-12-06
feat(useWallet): add correctNetwork flag

### LINKED ISSUE #72: [useWallet] Add support for wrong network detection
The `useWallet` hook does not have any variable or method that notifies the consuming component about whether the user is connected to the right network or not. Something like this would be nice:

```tsx
const { isNetworkCorrect }  = useWallet(); // returns true or false
```

Dhaiwat10:
Closes #72

## Description
This PR adds the `correctNetwork` flag to the `useWallet` hook  whose value changes between `true` -> `false` depending on whether the user is on the correct network supplied to the `<Provider />`.

## Other changes
- Added event listeners for network change, account change and account disconnect.
- Cleaned up the `useWallet` hook a bit
- The `<Provider />` now accepts the network's chain ID (number) instead of the network's name as a string. This lets people pass in a custom chain ID.
- To aid UX, I have added a constants file containing mappings between the network names and their respective chain IDs. Usage example: `<Provider network={NETWORKS.mainnet} />`


Dhaiwat10:
Sorry for introducing so many changes in one PR but they kinda depend on each other :(

packages/hooks/src/stories/ConnectWallet.stories.tsx
```diff
@@ -1,19 +1,26 @@
...
 
 export default {
   title: 'Hooks/useWallet',
 };
 
 const DefaultUsingProvider = () => {
-  const { connection, connectWallet, disconnectWallet, connected } = useWallet();
+  const { connection, connectWallet, disconnectWallet, connected, correctNetwork } = useWallet();
+
+  useEffect(() => {
```
christiananese, line 12: Any reason your using a named import here and we use the default React import everywhere else?

Just for my understanding, I personally find it confusing if both are used.

Dhaiwat10: Are you referring to using `useEffect` instead of `React.useEffect`?

Dhaiwat10: Oh yeah that's just me being dumb. 😬 I'll fix that

PR #88, opened 2021-12-08
set Storybook default viewmode to "Docs"

txwkx:
## Description

Set the `Docs` tab as a default view mode for stories in Storybook.

image

## 📝 Additional Information

⚠️ You might notice that the initial render of the Storybook has the `Canvas` tab open and only after clicking on a different story it would start defaulting to `Docs`. I've checked with my other Storybook projects that use this addon - same behaviour. 

Suboptimal, but quite minor tbf

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -8,3 +8,8 @@ export const decorators = [
     </ChakraProvider>
   ),
 ];
+
+export const parameters = {
+  controls: { expanded: true },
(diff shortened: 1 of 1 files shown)


etr2460 (approved):
lgtm. a bit disappointing this doesn't work in chromatic, but it's strictly an improvement, so lets merge

etr2460:
@allcontributors please add @txwkx for code

PR #90, opened 2021-12-09
EtherInput component

### LINKED ISSUE #10: Token amount/value input with easy conversion to-from wei<->ether
Please take inspiration from [scaffold-eth's `EtherInput` component](https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/react-app/src/components/EtherInput.jsx).

swetshaw:
Closes  #10 

## Description

- This component provides an input element with built-in Ether to Wei converter
- The component accepts `value` as a prop that can be converted into Wei or Ether as required
- By default, the selected unit is `Ether`

## 📝 Additional Information

- Right now the conversion logic (Ether <-> Wei) is written in the codebase itself
- We can also provide options for converting into other units (kwei, mwei, finney, etc.) as well

## ⏳ Pending task

- [ ] Writing tests for the component


packages/components/src/components/ConvertTokenInput/ConvertTokenInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (targetUnit === 'wei') {
+      setCurrentUnit('wei');
+      setTargetUnit('ether');
+      setVal((parseFloat(val) * 1000000000000000000).toFixed(2));
+    } else if (targetUnit === 'ether') {
+      setCurrentUnit('ether');
+      setTargetUnit('wei');
+      setVal((parseFloat(val) / 1000000000000000000).toFixed(18));
+    }
+  };
```
christiananese, line 38: Maybe we could move this to a custom hook? I think it's something we might reuse in other places as well. @Dhaiwat10 @etr2460

swetshaw: We definitely can @christiananese , but this functionality is already provided by [web3js](https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#fromwei) and [ether-converter package](https://www.npmjs.com/package/ether-converter) . Will it be a good idea to use these libraries?

packages/components/package.json
```diff
@@ -33,7 +33,8 @@
     "@emotion/styled": "^11",
     "classnames": "^2.2.6",
     "cross-fetch": "^3.1.4",
-    "framer-motion": "^4"
+    "framer-motion": "^4",
+    "web3": "^1.6.1"
```
Dhaiwat10, line 37: We already have `ethers`, can you use that instead?

packages/components/package.json
```diff
@@ -44,6 +45,7 @@
     "@storybook/react": "^6.3.12",
     "@types/classnames": "^2.2.11",
     "@types/jest": "^26.0.15",
+    "@types/mocha": "^9.0.0",
```
Dhaiwat10, line 48: I think this was added by mistake. We aren't using Mocha AFAIK. Can you remove this?

packages/components/src/components/ConvertTokenInput/ConvertTokenInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [currentUnit, setCurrentUnit] = useState('ether');
+  const [targetUnit, setTargetUnit] = useState('wei');
+  const [val, setVal] = useState(value);
+
+  // The convert() function converts the current value into the unit selected by the user
+  const convert = () => {
+    if (targetUnit === 'wei') {
+      setCurrentUnit('wei');
+      setTargetUnit('ether');
+      setVal(Web3.utils.toWei(val));
```
Dhaiwat10, line 33: Same as above. Use `ethers.utils`

Dhaiwat10 (changes requested):
Thanks! Left some comments. I am also marking this PR ready for review so that this gets deployed to Chromatic and I can have a look at the Storybook

packages/components/package.json
swetshaw: cool, am checking 🙂

packages/components/package.json
swetshaw: Yes, this happened while I trying to setup the test environment, I will remove this 👍

Dhaiwat10:
@swetshaw any updates on this?

packages/components/src/components/ConvertTokenInput/ConvertTokenInput.tsx
swetshaw: Done 👍

Dhaiwat10:
@swetshaw one problem I noticed is that when I try to pass in `width='30%'` to the component, only the inner input's width is getting affected and not that button on the right. This might be because you are spreading the `props` on the inner `Input` instead of spreading them over `InputGroup`

image

swetshaw:
> @swetshaw one problem I noticed is that when I try to pass in `width='30%'` to the component, only the inner input's width is getting affected and not that button on the right. This might be because you are spreading the `props` on the inner `Input` instead of spreading them over `InputGroup`
> 
> image

Hi @Dhaiwat10 , I have added `InputGroupProps` and spread it over InputGroup.

swetshaw:
@Dhaiwat10 , if you find any other issues or have any suggestions do let me know 🙌

Dhaiwat10:
https://gist.github.com/Dhaiwat10/7ecb5325587ef947deae3418f5225520

@swetshaw this gist should give you some idea

Also this screenshot could help: (ignore `placeholder`, not needed)

image

swetshaw:
Thanks @Dhaiwat10 , you have made it easy 🙂

Dhaiwat10:
@swetshaw thanks for continuing to work on this! If you don't have anything more to push, would you mind if I made some edits?

swetshaw:
> @swetshaw thanks for continuing to work on this! If you don't have anything more to push, would you mind if I made some edits?

Yes sure. I won't mind.

Dhaiwat10:
@allcontributors please add @swetshaw for code!

PR #91, opened 2021-12-09
ShortAddress component

joselvelez:
Feature: Short Address Component

## Description
A component that displays a shortened address if the address does not resolve to an ENS address.

image

### DIFF
--- a/packages/components/src/components/ShortAddress/ShortAddress.stories.tsx
+++ b/packages/components/src/components/ShortAddress/ShortAddress.stories.tsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import { ShortAddress } from '.';
+
+import { useWallet, Provider, NETWORKS } from '@web3-ui/hooks';
+import { Button } from '@chakra-ui/react';
+
+export default {
+  title: 'Components/ShortAddress',
+  component: ShortAddress,
+};
+
+// Default => no address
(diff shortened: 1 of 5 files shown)


etr2460 (commented):
A general comment, since a lot of these seems to be similar to the `Address` component, perhaps we should add a `short` or `compact` prop to the `Address` component that then renders the shortened address?

One other food for thought, it could be nice to have a copy address button when we display the shortened address, since there's no other way to get it then. probably doesn't need to be implemented in this PR though

joselvelez:
> 
Good idea on the prop. I was thinking about that and was not sure how to best do it. It would not be a hook and I considered perhaps saving the logic inside a callable function but was not sure where to store it? And I was not sure how I would get this working in storybook since it just had hooks and components. 

I agree, I wanted to add a copy address button as well. I decided to leave it out of this PR to give other folks like me that are new to making PRs to projects a chance get their feet wet. It has been immensely rewarding and helpful me. That and I'm not at all good with styling buttons! :)

txwkx:
I'll review the PR once a feature branch is available. 

@joselvelez close this PR after the new one gets created, please 🙏

joselvelez:
I am closing this PR and will submit a new one as a separate branch.

joselvelez:
I am closing this PR and will submit a new one as a separate branch.

Issue #86, opened 2021-12-07
Update Address component to display shortened address


maximebonhomme:
### Description

Would be handy to be able to display a shortened version of the address as well as having a one click copy to clipboard functionality on it.

### Why is this needed?

Wallet addresses are far too long and in most cases we don't need to display it in full in our dapp.

Showing the first 4 and last 4 chars is usually enough to get a glimpse of which address we're using.

We usually do things with an address, like copy and paste it into etherscan to check some transactions, paste it into opensea to see the account's NFTs etc... Since we're only showing the first/last 4 chars we need a way to copy to clipboard easily.

### Proposed Solution or API

This ticket includes 2 changes really:
- New prop on `Address` called `shortened` (bool) to display 4 first/last chars of an address
- Copy full address to clipboard onClick

Good to haves would be:
- Visual hint to show that the full address can be copied
- Visual feedback when address has been successfully copied to clipboard

### Additional Information

Let's start with the basic requirements and can work on the good to haves if we have time.

Make sure to create a new ticket with good to haves (as I think they are valuable) if we can't do this as part of this issue for some reason.

maximebonhomme:
Can anyone assign this to me @Dhaiwat10 ? Will aim to finish this by the 13th

Dhaiwat10:
Done @maximebonhomme, thanks for taking the initiative on this one!

joselvelez:
I just noticed this issue before submitting PR #91 with some of this functionality. Should I cancel my PR?

Dhaiwat10:
@joselvelez no keep it for now. Let's wait until @maximebonhomme turns in their PR

maximebonhomme:
amazing @joselvelez just saw your PR. I haven't started working on this yet but feel free to work on this if you want, otherwise I'll make a start on this tomorrow :)

joselvelez:
I just added #95 to add the shorten functionality. It does not have the copy address functionality yet. I'm doing small bits to get a hang of making contributions. Let me know if you want to do the copy address piece.

maximebonhomme:
amazing thanks @joselvelez! Let's get that PR reviewed and merged and I'll work on the copy address functionality today 👍

PR #95, opened 2021-12-11
update Address component with shortAddress prop

joselvelez:
Update the address component to address #86.

This is only handles the shortening aspect. Not the button to allow copying the address.
image

### DIFF
--- a/packages/components/src/components/Address/Address.stories.tsx
+++ b/packages/components/src/components/Address/Address.stories.tsx
@@ -9,7 +9,7 @@ export default {
   component: Address,
 };
 
-export const Default = () => <Address value='testaddress.eth' shortened={false} />;
+export const Default = () => <Address value='testaddress.eth' />;
 
 export const DefaultShortenedWithENS = () => <Address value='testaddress.eth' shortened={true} />;
 
--- a/packages/components/src/components/Address/Address.stories.tsx
+++ b/packages/components/src/components/Address/Address.stories.tsx
@@ -33,7 +33,7 @@ const AddressUsingProvider = ({ shortened }) => {
 
 export const WithWallet = () => (
   <Provider network={NETWORKS.rinkeby}>
-    <AddressUsingProvider shortened={false} />
+    <AddressUsingProvider />
   </Provider>
 );
 
--- a/packages/components/src/components/Address/Address.stories.tsx
+++ b/packages/components/src/components/Address/Address.stories.tsx
@@ -9,15 +9,22 @@ export default {
   component: Address,
 };
 
-export const Default = () => <Address value='testaddress.eth' />;
+export const Default = () => <Address value='testaddress.eth' shortened={false} />;
 
-const AddressUsingProvider = () => {
+export const DefaultShortenedWithENS = () => <Address value='testaddress.eth' shortened={true} />;
+
+export const DefaultShortenedWithHexAddress = () => (
+  <Address value='0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b' shortened={true} />
+);
+
+const AddressUsingProvider = ({ shortened }) => {
   const { connected, connectWallet, connection } = useWallet();
 
   return (
     <>
       <Address
         value={connected ? connection.ens || connection.userAddress || '' : 'Not connected'}
+        shortened={shortened}
       />
       <Button onClick={connectWallet}>Connect wallet</Button>
     </>
@@ -26,6 +33,12 @@ const AddressUsingProvider = () => {
 
 export const WithWallet = () => (
   <Provider network={NETWORKS.rinkeby}>
-    <AddressUsingProvider />
+    <AddressUsingProvider shortened={false} />
+  </Provider>
+);
+
+export const WithWalletShortened = () => (
+  <Provider network={NETWORKS.rinkeby}>
+    <AddressUsingProvider shortened={true} />
   </Provider>
 );
--- a/packages/components/src/components/Address/Address.stories.tsx
+++ b/packages/components/src/components/Address/Address.stories.tsx
@@ -39,6 +39,6 @@ export const WithWallet = () => (
 
 export const WithWalletShortened = () => (
   <Provider network={NETWORKS.rinkeby}>
-    <AddressUsingProvider shortened={true} />
+    <AddressUsingProvider shortened />
   </Provider>
 );
--- a/packages/components/src/components/Address/Address.stories.tsx
+++ b/packages/components/src/components/Address/Address.stories.tsx
@@ -14,7 +14,7 @@ export const Default = () => <Address value='testaddress.eth' />;
 export const DefaultShortenedWithENS = () => <Address shortened value='testaddress.eth' />;
 
 export const DefaultShortenedWithHexAddress = () => (
-  <Address value='0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b' shortened={true} />
+  <Address shortened value='0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b' />
 );
 
 const AddressUsingProvider = ({ shortened }) => {
--- a/packages/components/src/components/Address/Address.stories.tsx
+++ b/packages/components/src/components/Address/Address.stories.tsx
@@ -11,7 +11,7 @@ export default {
 
 export const Default = () => <Address value='testaddress.eth' />;
 
-export const DefaultShortenedWithENS = () => <Address value='testaddress.eth' shortened={true} />;
+export const DefaultShortenedWithENS = () => <Address shortened value='testaddress.eth' />;
 
 export const DefaultShortenedWithHexAddress = () => (
   <Address value='0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b' shortened={true} />
--- a/packages/components/src/components/Address/Address.stories.tsx
+++ b/packages/components/src/components/Address/Address.stories.tsx
@@ -17,14 +17,18 @@ export const DefaultShortenedWithHexAddress = () => (
   <Address shortened value='0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b' />
 );
 
-const AddressUsingProvider = ({ shortened }) => {
+type AddressProps = {
+  shortened?: boolean;
+};
+
+const AddressUsingProvider = (props: AddressProps) => {
   const { connected, connectWallet, connection } = useWallet();
 
   return (
     <>
       <Address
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Address/Address.test.tsx
```diff
@@ -5,7 +5,7 @@ import { Address } from './Address';
 
 describe('Address', () => {
   it('renders without throwing', () => {
-    const { container } = render(<Address value="taylorswift.eth" />);
+    const { container } = render(<Address value='taylorswift.eth' shortened={false} />);
     expect(container).toBeInTheDocument();
   });
 });
```
txwkx, line 11: It's worth adding a test to check the output length of the shortened version

packages/components/src/components/Address/Address.stories.tsx
```diff
@@ -26,6 +33,12 @@ const AddressUsingProvider = () => {
...
 export const WithWallet = () => (
   <Provider network={NETWORKS.rinkeby}>
-    <AddressUsingProvider />
+    <AddressUsingProvider shortened={false} />
+  </Provider>
+);
+
+export const WithWalletShortened = () => (
+  <Provider network={NETWORKS.rinkeby}>
+    <AddressUsingProvider shortened={true} />
```
txwkx, line 42:
```suggestion
    <AddressUsingProvider shortened />
```

Omitting the value of the prop when it is explicitly `true` is recommended

packages/components/src/components/Address/Address.tsx
```diff
@@ -6,11 +6,27 @@ export interface AddressProps {
...
-  return <Input value={value} />;
+export const Address: React.FC<AddressProps> = ({ value, shortened }) => {
+  if (shortened) {
+    if (value.includes('.eth')) {
+      return <Input value={value} />;
+    } else if (value === '' || value === 'Not connected') {
+      return <Input value={value} />;
+    } else {
+      let address: string;
+      address = `${value.substring(0, 4)}...${value.substring(value.length - 4)}`;
```
txwkx, line 26: minor one, but is there a need for the let variable? `const address = ...` should work

packages/components/src/components/Address/Address.tsx
```diff
@@ -6,11 +6,27 @@ export interface AddressProps {
...
 /**
  * A component to display an address
  */
-export const Address: React.FC<AddressProps> = ({ value }) => {
-  return <Input value={value} />;
+export const Address: React.FC<AddressProps> = ({ value, shortened }) => {
+  if (shortened) {
+    if (value.includes('.eth')) {
+      return <Input value={value} />;
+    } else if (value === '' || value === 'Not connected') {
```
txwkx, line 22: If I understand correctly you are accounting for an empty instance here.

Perhaps it would be easier to drop this `if statement` in favour of a `placeholder` prop on the `<Input />` instead?

https://chakra-ui.com/docs/form/input#usage

packages/components/src/components/Address/Address.tsx
```diff
@@ -6,11 +6,27 @@ export interface AddressProps {
...
+   */
+  shortened: boolean;
 }
 
 /**
  * A component to display an address
  */
-export const Address: React.FC<AddressProps> = ({ value }) => {
-  return <Input value={value} />;
+export const Address: React.FC<AddressProps> = ({ value, shortened }) => {
```
txwkx, line 18: add `import React, { FC } from 'react';` to drop the React here

packages/components/src/components/Address/Address.tsx
```diff
@@ -6,11 +6,27 @@ export interface AddressProps {
...
+      return <Input value={value} />;
+    } else if (value === '' || value === 'Not connected') {
+      return <Input value={value} />;
+    } else {
+      let address: string;
+      address = `${value.substring(0, 4)}...${value.substring(value.length - 4)}`;
+      return <Input value={address.toLowerCase()} />;
+    }
+  } else {
+    return <Input value={value} />;
```
txwkx, line 30: on a fundamental level, it may be worth having a single `return <Input value={%_var_%} />;` at the end of this component. And focus the conditional logic to assign a respective value to a `let` variable

e.g.

```js
let displayAddress;

if(case) {
 displayAddress = 'a';
} else if (case2) {
 displayAddress = 'b';
}

return <Input value={displayAddress} />

**Review by txwkx — CHANGES_REQUESTED** [review]
comments left 🙏

File: packages/components/src/components/Address/Address.tsx [hunk 7]
```diff
@@ -6,11 +6,27 @@ export interface AddressProps {
    * The address to display
    */
   value: string;
+  /**
+   * boolean to shorten the address if addresses does not resolve to an ENS name
```
etr2460, line 10:
```suggestion
   * Shorten the address if it does not resolve to an ENS name
```

etr2460: no need to note that it's a boolean, we'll get that from the type automatically

packages/components/src/components/Address/Address.tsx
```diff
@@ -6,11 +6,27 @@ export interface AddressProps {
...
+   */
+  shortened: boolean;
```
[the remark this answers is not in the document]
etr2460: ++ on making it optional. We actually don't need a default value because `undefined` is falsey
```suggestion
  shortened?: boolean;
```

packages/components/src/components/Address/Address.test.tsx
```diff
@@ -5,7 +5,7 @@ import { Address } from './Address';
 
 describe('Address', () => {
   it('renders without throwing', () => {
-    const { container } = render(<Address value="taylorswift.eth" />);
+    const { container } = render(<Address value='taylorswift.eth' shortened={false} />);
```
etr2460, line 8: You also won't need to set shortened={false} once you make that prop optional

```suggestion
    const { container } = render(<Address value='taylorswift.eth' />);
```

packages/components/src/components/Address/Address.stories.tsx
```diff
@@ -9,15 +9,22 @@ export default {
   component: Address,
 };
 
-export const Default = () => <Address value='testaddress.eth' />;
+export const Default = () => <Address value='testaddress.eth' shortened={false} />;
```
etr2460, line 12:
```suggestion
export const Default = () => <Address value='testaddress.eth' />;
```

packages/components/src/components/Address/Address.stories.tsx
```diff
@@ -9,15 +9,22 @@ export default {
...
 
-const AddressUsingProvider = () => {
+export const DefaultShortenedWithENS = () => <Address value='testaddress.eth' shortened={true} />;
```
etr2460, line 14: true is implied when you set a boolean prop on a component in this way

```suggestion
export const DefaultShortenedWithENS = () => <Address shortened value='testaddress.eth' />;
```

packages/components/src/components/Address/Address.stories.tsx
```diff
@@ -9,15 +9,22 @@ export default {
...
+
+export const DefaultShortenedWithHexAddress = () => (
+  <Address value='0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b' shortened={true} />
```
etr2460, line 17:
```suggestion
  <Address shortened value='0x7Be8076f4EA4A4AD08075C2508e481d6C946D12b' />
```

packages/components/src/components/Address/Address.stories.tsx
```diff
@@ -26,6 +33,12 @@ const AddressUsingProvider = () => {
 
 export const WithWallet = () => (
   <Provider network={NETWORKS.rinkeby}>
-    <AddressUsingProvider />
+    <AddressUsingProvider shortened={false} />
```
etr2460, line 36:
```suggestion
    <AddressUsingProvider />
```

etr2460 (commented):
looking great! The majority of the comments I have are around setting the boolean prop. I think this will be good to go soon though, thanks for all the hard work!

packages/components/src/components/Address/Address.tsx
maximebonhomme: Could we set the `shortened` default value to `false` for readability? it will also display default to false in Storybook

```jsx
({ value, shortened = false })
```

packages/components/src/components/Address/Address.tsx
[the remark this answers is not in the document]
joselvelez: That's a good point!

packages/components/src/components/Address/Address.stories.tsx
joselvelez: Ah I see, since it resolves to true it is not necessary to be explicit about it?

packages/components/src/components/Address/Address.tsx
joselvelez: made both changes. Thanks!

packages/components/src/components/Address/Address.tsx
joselvelez: I added a check for 'Not connected' because if that value gets passed into the prop it displays the shortened version of it.

C:\Users\power\Downloads\2021-12-12 16_34_14-Components _ Address - Default ⋅ Storybook - Brave.png

If we went with placeholder prop, would that be in addition to passing the value? so `value={value} placeholder={value}`?

packages/components/src/components/Address/Address.test.tsx
```diff
@@ -5,7 +5,15 @@ import { Address } from './Address';
...
   });
 });
+
+describe('Shortened Address', () => {
+  it('check the length of the address', () => {
+    const { container } = render(<Address value='0x00000000000000' shortened />);
+    const addressInput = container.querySelector('input') as HTMLInputElement;
+    expect(addressInput).toHaveValue('0x00...0000');
+  });
+});
```
etr2460, line 19:
```suggestion
    const { container } = render(<Address value='taylorswift.eth' shortened={false} />);
    expect(container).toBeInTheDocument();
  });
  
  it('checks the length of the address when shortened', () => {
    const { container } = render(<Address value='0x00000000000000' shortened />);
    const addressInput = container.querySelector('input') as HTMLInputElement;
    expect(addressInput).toHaveValue('0x00...0000');
  });
});
```

etr2460 (approved):
one tiny nit about how the test is setup, otherwise this looks awesome! thanks for all the hard work and iteration here

etr2460:
@allcontributors add @joselvelez for code

PR #97, opened 2021-12-12
TokenBalance Component 

### LINKED ISSUE #42: Token Balance Component
Hi, I'm thinking to build a component, that is displaying the user balance of a certain token. 
Personally, I like the approach dotmerketcap.com goes design-wise.

So I think the component would have the following params

- Contract Address (the address where the token's contract is deployed)
- Img (Path to Logo)
- Value in USD (the ability to pass the current price of the token per USD )

If the user is connected with his wallet then the balance of the certain token will be fetched. Otherwise, I think we can show them either an empty state or just the balance of 0.

Screenshot 2021-11-30 at 23 06 50

AlexNi245:
Closes #42 

## Description

This component shows the amount of a certain ERC20 token that the connected user owns. Besides displaying it has the following features:
* Fetching the owners balance of the selected token using the useTokenBalance hook
* Calculation of the USD value of the balance by multiplying balance * provided USD/token rate
* displaying the logos logo by using the trustwallet github repo
* Fetching symbol and name of the token from the according smart contract

## 📝 Additional Information

During the work i noticed a few things. I would love to hear your opinions on this @Dhaiwat10 @etr2460 @Ibby-devv 

* Are we are using some common empty / error state across the library ?  Currently, the behavior of the component on error is to display the balance of 0
* I decided to not use APIs such as coinmarketcap to fetch the USD value because I think most users want to provide their own exchange rate. But I would be happy to discuss this. 
* I also decided to just display the first 3 decimal places of the value and the balance. 
* I think I could make sense to pass name and symbol as a prop rather than fetch these values from the contract. This would reduce loading times (during loading name and symbol are just empty string) and would display name and symbol anyway even there are problems with the web3 Provider. 

## Screenshot
Screenshot 2021-12-12 at 21 08 07


Dhaiwat10:
@AlexNi245 sorry for not getting back to this yet. Since this PR has quite a bit of code I just need a bit more to review everything. Thanks for your patience mate

.changeset/config.json
```diff
@@ -1,12 +1,9 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@1.6.3/schema.json",
-  "changelog": [
```
JoviDeCroock, line 3: Why are you removing this, this is repo config

package.json
```diff
@@ -46,6 +46,12 @@
     "msw": "^0.35.0",
     "ts-jest": "^26.4.4"
   },
+  "dependencies": {
+    "@web3-ui/hooks": "0.4.0"
+  },
+  "peerDependencies": {
```
JoviDeCroock, line 52: It's already in the regular deps, we put it here only if we require the user to manually install the package in question

packages/components/src/components/TokenBalance/TokenBalance.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const contract = new ethers.Contract(tokenAddress, ERC20ABI, provider!);
+    const name = await contract.name();
+    const symbol = await contract.symbol();
+
+    setName(name);
+    setSymbol(symbol);
+  };
+
+  const Headline = (name: string) => (
+    <Heading as='h3' size='sm'>
```
JoviDeCroock, line 66: Inline components can't maintain their referential identity and are subject to constant remounting. Hoisting outside of this closure will resolve that

JoviDeCroock (changes requested):

package.json
AlexNi245: I tried to solve the failing Ci step by adding this. Unfortunately, it doesn't solve that issue. Have you an idea what to do?

.changeset/config.json
AlexNi245: Actually, this was not on purpose. Must have occurred during the upstream merge

package.json
JoviDeCroock: Because you are importing from `src`

.changeset/config.json
AlexNi245: I'll revert this

packages/components/src/components/TokenBalance/TokenBalance.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { useContext, useEffect, useState } from 'react';
+import { Box, Flex, Heading, Image } from '@chakra-ui/react';
+import { TokenLogo } from './components/Logo';
+import { useTokenBalance, Web3Context } from '@web3-ui/hooks/src';
```
JoviDeCroock, line 4: @AlexNi245 cross package imports like these are _not_ allowed in libs

JoviDeCroock: You still have some

```
error @web3-ui/components "@web3-ui/hooks" is imported by "src/components/TokenBalance/TokenBalance.tsx" but the package is not specified in dependencies or peerDependencies
```

JoviDeCroock:
Your linter seems a bit goofy 😅 you will have to add the package here https://github.com/Developer-DAO/web3-ui/blob/main/packages/components/package.json#L30

JoviDeCroock:
Just need to add a changeset "minor" and we're there, you can use the link here https://github.com/Developer-DAO/web3-ui/pull/97#issuecomment-992923169 or use `yarn changeset` at the repo root

packages/components/src/components/TokenBalance/components/Logo.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { Image } from '@chakra-ui/react';
+
+export const TokenLogo = ({ tokenContractAddress }) => (
+  <Image
+    src={`https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/${tokenContractAddress}/logo.png`}
+  />
+);
+
+//i.E. Dai
+//https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0x6B175474E89094C44Da98b954EedeAC495271d0F/logo.png
```
Dhaiwat10, line 11: Can we move this component over to the `TokenBalance.tsx`? We don't need a separate file for this

packages/hooks/src/Provider.tsx
```diff
@@ -89,7 +89,7 @@ export const Provider: React.FC<ProviderProps> = ({
     setConnection(connection);
     const provider = new ethers.providers.Web3Provider(connection);
     setProvider(provider);
-    const chainId = await provider.getNetwork().then(network => network.chainId);
+    const chainId = await provider.getNetwork().then((network) => network.chainId);
```
Dhaiwat10, line 92: You should revert this since I don't think this is in the scope of this PR

packages/hooks/src/Provider.tsx
```diff
@@ -126,7 +126,7 @@ export const Provider: React.FC<ProviderProps> = ({
   const onAccountsChanged = async () => {
     const provider = new ethers.providers.Web3Provider(connection);
     setProvider(provider);
-    const chainId = await provider.getNetwork().then(network => network.chainId);
+    const chainId = await provider.getNetwork().then((network) => network.chainId);
```
Dhaiwat10, line 129: Same as above!

packages/components/src/components/TokenBalance/components/Headline.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Heading } from '@chakra-ui/react';
+
+export const Headline = (name: string) => (
+  <Heading as='h3' size='sm'>
+    {name}
+  </Heading>
+);
```
Dhaiwat10, line 8: Can we also move this to `TokenBalance.tsx`?

packages/components/src/components/TokenBalance/TokenBalance.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const getTokenNameAndSymbol = async () => {
+    const contract = new ethers.Contract(tokenAddress, ERC20ABI, provider!);
+    const [name, symbol] = await Promise.all([contract.name(), contract.symbol()]);
+
+    setName(name);
+    setSymbol(symbol);
+  };
+
+  return (
+    <Flex borderRadius='lg' borderWidth='1px' overflow='hidden' px='4' py='2'>
```
Dhaiwat10, line 66: Defining static styles and not letting the user set these styles if they want is not what we want. Can we accept all the props a Chakra `Flex` component would expect and spread them here? Our `<AddressInput />` does the same for the Chakra `Input` component.

.changeset/config.json
```diff
@@ -2,11 +2,13 @@
   "$schema": "https://unpkg.com/@changesets/config@1.6.3/schema.json",
   "changelog": [
     "@changesets/changelog-github",
-    { "repo": "Developer-DAO/web3-ui" }
+    {
+      "repo": "Developer-DAO/web3-ui"
+    }
```
Dhaiwat10, line 7: Can you revert this, please?

packages/components/src/components/TokenBalance/components/Logo.tsx
Dhaiwat10: You can also remove those comments

Dhaiwat10 (changes requested):
Left some comments!

packages/components/src/components/TokenBalance/TokenBalance.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    setName(name);
+    setSymbol(symbol);
+  };
+
+  const Headline = (name: string) => (
+    <Heading as='h3' size='sm'>
+      {name}
+    </Heading>
+  );
```
Dhaiwat10, line 66: Like Jovi suggested:

```suggestion
  const Headline = (name: string) => { 
    return (
      <Heading as='h3' size='sm'>
        {name}
      </Heading>
    );
  }
```

.changeset/config.json
```diff
@@ -1,12 +1,9 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@1.6.3/schema.json",
-  "changelog": [
-    "@changesets/changelog-github",
-    { "repo": "Developer-DAO/web3-ui" }
-  ],
+  "changelog": ["@changesets/changelog-github", { "repo": "Developer-DAO/web3-ui" }],
```
Dhaiwat10, line 3: This should not be changed. Can you make sure this change goes away please?

packages/components/src/components/TokenBalance/erc.json
```diff
@@ -0,0 +1,222 @@
...
+    "name": "name",
+    "outputs": [
+      {
+        "name": "",
+        "type": "string"
+      }
+    ],
+    "payable": false,
+    "stateMutability": "view",
+    "type": "function"
```
Dhaiwat10, line 14: Can you rename this file to `ERC20.json` instead?

packages/hooks/src/index.ts
```diff
@@ -1,5 +1,5 @@
-export { useWallet, useContract } from './hooks';
-export { useTokenBalance } from './hooks/useTokenBalance';
-export { Provider, Web3Context } from './Provider';
-export type { ProviderProps, Web3ContextType } from './Provider';
-export { NETWORKS, CHAIN_ID_TO_NETWORK } from './constants';
+export {useWallet, useContract} from './hooks';
+export {useTokenBalance} from './hooks/useTokenBalance';
+export {Provider, Web3Context} from './Provider';
+export type {ProviderProps, Web3ContextType} from './Provider';
+export {NETWORKS, CHAIN_ID_TO_NETWORK, ERC20ABI} from './constants';
```
Dhaiwat10, line 5: All of these lines should not be changed. Only the last line is actually being changed. Can you reset these?

Dhaiwat10 (changes requested):
Can you make these small changes please and we're good to go

.changeset/config.json
AlexNi245: sir, I almost lost my mind on this but I got the feeling this file was formatted by the commit hook.  I have pushed it multiple times just to notice that this file still has unwanted changes.

Dhaiwat10:
@AlexNi245 I think you commited some `.idea` files by mistake

packages/hooks/src/index.ts
AlexNi245: This must be also the result of the post commit hook. I reverted the first few lines just to notice that they have changed again after I committed my changes. 
I can use -no--verify again to restore the first 5 lines. But maybe it's intended that the commit hook applies certain formatting rules to each file even the affected file has nothing to do with the pr

packages/components/package.json
```diff
@@ -28,6 +28,7 @@
   "module": "dist/web3-ui-components.esm.js",
   "types": "dist/web3-ui-components.cjs.d.ts",
   "dependencies": {
+    "@web3-ui/hooks": "0.4.0",
```
Dhaiwat10, line 31: One more thing that came to my mind recently is that the `components` package should not rely on the `hooks` package in any way. We want both of these packages to function well together but also independently. 

This is why I feel like we shouldn't do any data fetching at all inside any of our `components`. Here is what I propose:

1. Move the current implementation to the `core` package.
2. Make the current implementation (inside `components`) so that it does not handle any data fetching but just accepts the data it needs to render.

Let me know if this is unclear. More than happy to discuss this in further detail since I think this is an important topic.

packages/hooks/src/index.ts
Dhaiwat10: I see. No worries then.

packages/components/package.json
AlexNi245: Do you mean this component shouldn't have any state at all and just renders given props ?  Then we can remove the useTokenBalance hook from the Component. 
Is that what you meant in the second point?

AlexNi245:
> @AlexNi245 I think you commited some `.idea` files by mistake

Sorry for that, they are deleted now

packages/components/package.json
Dhaiwat10: @AlexNi245 yes! The current verison of this component which _does_ have some state and fetches data should be in the `core` package and not in the `components` package. This is what we did recently for #121 too.

tldr:

- `<TokenBalance />` in `components` shouldn't have any state, it should just render some data that you pass to it.
- `<TokenBalance />` in `core` **should** have state and should use the `useTokenBalance` hook

Dhaiwat10:
@AlexNi245 just checking in. Do you need any help?

AlexNi245:
Hi @Dhaiwat10 actually I was on vacation since Christmas.I will come back home on Thursday. My apologies for not responding but I will apply your suggestions till the end of the week:)

Dhaiwat10:
No worries at all mate, enjoy!

Dhaiwat10:
@AlexNi245 since this PR's history was extremely messed up, I copied your changes from here and opened a new PR here: #237. Closing this one

PR #105, opened 2021-12-13
Feat: address can be copied to clipboard

### LINKED ISSUE #86: Update Address component to display shortened address

maximebonhomme:
Let's wait for PR#95 to be merged first and I'll fix any conflicts

Closes #86 

## Description

This PR includes changes to make the Address component copiable to clipboard. We're now able to provide a prop `<Address value='0x000' copiable />` to enable users to copy the address to their clipboard.

## What changed

- New prop on `Address` called `copiable` as type `boolean` - default is falsy
- When `copiable` is set to `true` we're now displaying a Copy Icon on the far right of the input - This allows users to know they can copy the address to their clipboard
- When users successfully copied the address to their clipboard we're replacing the Copy Icon by a Check Icon as a success feedback
- Added error state to notify users if copy to clipboard has failed

## 📝 Additional Information

I have added `@chakra-ui/icons` package to add icons, thinking that's the most suitable for us atm as we're already using chakraUI - any concerns let me know

## Screenshot

copiable set to true:
copiable set to true

copied successfully:
copied


maximebonhomme:
No ready to review - will resolve conflicts tomorrow

packages/components/src/components/Address/Address.test.tsx
```diff
@@ -1,11 +1,45 @@
...
+import { jest } from '@jest/globals';
+import { render, fireEvent, waitFor } from '@testing-library/react';
 
 import { Address } from './Address';
 
+Object.assign(navigator, {
+  clipboard: {
+    writeText: () => {},
+  },
+});
```
Dhaiwat10, line 11: Can you add a comment here that explains what this code does? (More importantly, why it was added?)

packages/components/src/components/Address/Address.tsx
```diff
@@ -1,16 +1,67 @@
...
-export const Address: React.FC<AddressProps> = ({ value }) => {
-  return <Input value={value} />;
+export const Address: React.FC<AddressProps> = ({ value, copiable = false }) => {
+  const [error, setError] = useState<null | string>(null);
+  const [copied, setCopied] = useState<boolean>(false);
+
+  const handleClick = async (event: SyntheticEvent): Promise<void> => {
+    const value = event.currentTarget.value as string;
+
+    setError(null);
```
Dhaiwat10, line 40: Why are we setting `error` to `null`?

packages/components/src/components/Address/Address.tsx
```diff
@@ -1,16 +1,67 @@
...
+    setCopied(false);
```
Dhaiwat10, line 41: I don't think this is necessary since it `copied` will be `true` almost instantly anyways later on in this very function.

What I suggest is that we set a timeout or something that sets `copied` back to `false` after 2 or 3 seconds so that the user can see the 'copy' icon after 2-3 seconds if they want to copy it again.

Dhaiwat10 (changes requested):
Thanks for the PR!

I also noticed that the cursor doesn't change to type `hover` when I hover over the copy button. Can you fix that please?

packages/components/src/components/Address/Address.test.tsx
maximebonhomme: fixed in [c12b761958d8a86a19c1220aa4b97d2aeff9ebaf](https://github.com/Developer-DAO/web3-ui/pull/105/commits/f14a7fb908e032e913e8dee684793cdb510c1b6d)

packages/components/src/components/Address/Address.tsx
maximebonhomme: fixed as part of https://github.com/Developer-DAO/web3-ui/pull/105/commits/dcd5eb650f1a3e4ad0cc5c9a2479ca1894c78cef

packages/components/src/components/Address/Address.tsx
```diff
@@ -1,16 +1,67 @@
...
+import { CopyIcon, CheckIcon } from '@chakra-ui/icons';
+import React, { useState } from 'react';
 
 export interface AddressProps {
   /**
    * The address to display
    */
   value: string;
+  /**
+   * Wheter the address can be copied or not
```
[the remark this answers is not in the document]
maximebonhomme: fixed as part of https://github.com/Developer-DAO/web3-ui/pull/105/commits/dcd5eb650f1a3e4ad0cc5c9a2479ca1894c78cef

packages/components/src/components/Address/Address.tsx
```diff
@@ -32,5 +54,40 @@ export const Address: FC<AddressProps> = ({ value, shortened = false }) => {
     displayAddress = value;
   }
 
-  return <Input value={displayAddress} />;
+  const handleClick = async (event: SyntheticEvent): Promise<void> => {
+    const value = event.currentTarget.value as string;
```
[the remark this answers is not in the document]
maximebonhomme: fixed in https://github.com/Developer-DAO/web3-ui/pull/105/commits/f16b638a0c482dd1f0200dc4263bd8066d4c8851

maximebonhomme:
@Dhaiwat10 added! Will create a ticket to convert that input into a text component as mentionned in Discord.

PR #116, opened 2021-12-13
include regenerator runtime babel runtime vars

### LINKED ISSUE #115: Don't include `.dev.js` stuff in the build
image

https://www.runpkg.com/?@web3-ui/core@0.1.0/dist/web3-ui-core.cjs.dev.js

This shouldn't be published to npm because when `process.env.NODE_ENV` is not `production`, the `dev.js` file is being used. 

```ts
'use strict';

if (process.env.NODE_ENV === "production") {
  module.exports = require("./web3-ui-core.cjs.prod.js");
} else {
  module.exports = require("./web3-ui-core.cjs.dev.js");
}
```

JoviDeCroock:
Closes #115

## Description

preconstruct doesn't automatically include `regeneratorRuntime` this fixes that crash

### DIFF
--- a/.changeset/gold-phones-unite.md
+++ b/.changeset/gold-phones-unite.md
@@ -1,0 +1,7 @@
+---
+'@web3-ui/components': patch
+'@web3-ui/core': patch
+'@web3-ui/hooks': patch
+---
+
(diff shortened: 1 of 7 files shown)


Dhaiwat10:
Awesome mate

PR #121, opened 2021-12-14
Token gate

### LINKED ISSUE #98: TokenGate component
### Description

I want to create a component that restricts access to all child components unless certain erc-20/erc-721 token requirements are met. It would take in a contract address and required number of tokens(defaults to 1) as props to determine access.

### Why is this needed?

I think adding this component would make it easier to restrict access to a small part or whole sections of UI based on different tokens or amounts of tokens in a users wallet. For example some UI elements might be unlocked at 1 token while others may require 10 tokens to gain access. Further parts of the UI could be locked by a completely different token. This wrapper component will make implementing something like that quick and easy.

### Proposed Solution or API

https://gist.github.com/hone1er/bff36ccb8b23f7c24fa9f93389272e3f

@Dhaiwat10 check it out

hone1er:
Closes #98 

## Description

This component restricts access to all child components unless certain erc-20/erc-721 token requirements are met. It takes as parameters a provider and contract address, and required number of tokens(defaults to 1) as props to determine access.

## 📝 Additional Information

Definitely would like some feedback on what should be returned by default in a few different cases. For example with no wallet connected it currently will return null by default unless a "deniedMessage" is provided.

Screen Shot 2021-12-14 at 8 18 03 AM


AlexNi245:
You should merge the current main branch into this one to resolve the merge conflicts

packages/components/src/components/TokenGate/TokenGate.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  return (
+    <>
+      <TokenGate
+        provider={provider!}
+        requiredQuantity={props.requiredQuantity}
+        contractAddress='0x25ed58c027921e14d86380ea2646e3a1b5c55a8b'
+      >
+        <Text>
+          You were able to access this component because you hold at least{' '}
```
AlexNi245, line 34: I would rather use a template string for adding the empty spaces.

packages/components/src/components/TokenGate/TokenGate.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+          {props.requiredQuantity == 0 ? 0 : props.requiredQuantity ? props.requiredQuantity : 1}{' '}
```
AlexNi245, line 35: As far as i understand this statement correct you try to check wether requiredQuantity is undefined? I think you can be simply this statement to something like this:

const text = requiredQuantity === undefined ? 0 : requiredQuantity

packages/components/src/components/TokenGate/TokenGate.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </WalletContextTestWrapper>
+    </Provider>
+  );
+};
+
+export const WithLoader = () => <Component loader={true} label='Loader' />;
+
+export const AccessDeniedDefault = () => <Component requiredQuantity={10000} label='Denied' />;
+
+const DeniedAccess = ({ requiredQuantity }) => (
```
AlexNi245, line 81: I think we should allow the user to pass his own child component that renders the error message or whatever he wants to display.

packages/components/src/components/TokenGate/TokenGate.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+export const TokenGate: React.FC<TokenGateProps> = ({
+  contractAddress,
+  provider,
+  requiredQuantity = 1,
+  children,
+  deniedMessage,
+  loader = false,
+}) => {
+  const erc20Abi = ['function balanceOf(address owner) view returns (uint256)'];
```
AlexNi245, line 35: you can import this from the constants file of this library 
import {ERC20ABI} from '@web3-ui/hooks';

packages/components/src/components/TokenGate/TokenGate.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [tokenQuantity, setTokenQuantity] = useState<number>(0);
+  const [loadedStatus, setloadedStatus] = useState<boolean>(false);
+  // connect to contract address to get balance
+  async function getTokenBalance() {
```
AlexNi245, line 39: u can use the useTokenBalance hook from this library aswell
  const { balance, decimals ,loading,error} = useTokenBalance({ tokenAddress, accountAddress });

packages/components/src/components/TokenGate/TokenGate.tsx
```diff
@@ -0,0 +1,80 @@
...
+      setloadedStatus(!loadedStatus);
+    }
+
+    return;
+  }
+
+  useEffect(() => {
+    async function setBalance() {
+      const balance = await getTokenBalance();
+      setTokenQuantity(parseInt(balance, 16));
```
AlexNi245, line 59: useTokenBalance also provides a field formattedBalance to display a user-readable and correctly calculated balance

packages/components/src/components/TokenGate/TokenGate.stories.tsx
hone1er: That's way better. It actually defaults to 1 so i'll do `const text = requiredQuantity === undefined ? 1 : requiredQuantity` but that should work.

packages/components/src/components/TokenGate/TokenGate.tsx
hone1er: Maybe i'll have to merge the main into my branch first cause i get an error that it doesn't exist.

packages/components/src/components/TokenGate/TokenGate.tsx
hone1er: So i was thinking more about the component a little more. Since it's simply a UI component should it only take the contract and balance as parameters? Then in the core repo wrap it with a provider for token balances? I was hesitant to use one of our hooks for the UI repo. Seemed more appropriate for a core component. Thoughts?

packages/components/src/components/Address/Address.test.tsx
```diff
@@ -6,7 +6,7 @@ import { Address } from './Address';
 
 /**
  * We need to mock the Clipboard API by creating a global navigator object.
- * 
+ *
```
Dhaiwat10, line 9: Can you revert all the changes in this file? I think it's not related to your PR. Let's keep the Git history and blame clean!

packages/components/src/components/TokenGate/TokenGate.tsx
Dhaiwat10: @hone1er yeah that's right. We don't want to add our `hooks` package as a dependency to the `components` / UI package.

packages/components/src/components/TokenGate/TokenGate.tsx
```diff
@@ -0,0 +1,83 @@
...
+  requiredQuantity = 1,
+  children,
+  deniedMessage,
+  loader = false,
+}) => {
+  const [tokenQuantity, setTokenQuantity] = useState<number>(0);
+  const [loadedStatus, setloadedStatus] = useState<boolean>(false);
+  // connect to contract address to get balance
+  async function getTokenBalance() {
+    const signer = provider!.getSigner();
```
Dhaiwat10, line 43: 1. Do we need a signer to read data?
2. If we _do_ need a signer in all cases, what do you think of accepting a `signer` as a prop instead of a `provider`?

packages/components/src/components/TokenGate/TokenGate.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+ */
+const DeniedAccess = props => (
+  <div>
+    <h1>This is a custom component for when access is denied</h1>
+    <ul>
+      <li>Make sure your wallet is connected</li>
+      <li>Verify you are connected to the correct address</li>
+      <li>
+        {`Make sure you hold the number of tokens required to access this component: 
+        ${props.requiredQuantity && props.requiredQuantity}`}
```
Dhaiwat10, line 77: It is coming up as `undefined` for me. :(
image

Dhaiwat10 (changes requested):
Thanks for the PR! Left a few questions.

image

Also, for some reason I cannot see these two stories

packages/components/src/components/TokenGate/TokenGate.tsx
hone1er: That would work. But should we take the signer at all? Or should we allow the dev to pass the quantity held by the wallet? 

In the core package it would use then use the get balance hook to provide that for the dev. 

Maybe the dev has the wallet balance in their app's state already so fetching the balance for them could be undesirable.

And if they want the balance fetched for them they can use the core package version.

hone1er:
> Thanks for the PR! Left a few questions.
> 
> image
> 
> Also, for some reason I cannot see these two stories



> Thanks for the PR! Left a few questions.
> 
> image
> 
> Also, for some reason I cannot see these two stories

On these it is because the default returns null if access is denied. So, access denied default should always be empty, and with loader should show a spinner while loading then show reactNode if you get access or null(the default) if you don't. With loader requires you to hold at least 1 DevDao NFT in the wallet you connected.

packages/components/src/components/TokenGate/TokenGate.tsx
Dhaiwat10: > In the core package it would use the get balance hook to provide that for the dev. Maybe the dev has the wallet balance in their app's state already so fetching the balance for them could be undesirable.

With you on this. Let's just take the balance as a prop :)

Dhaiwat10: @hone1er do you also want to add Token Gate to `core` in this PR?

hone1er (commented):
I removed much of the logic. Now it just takes the walletBalance instead of taking a provider and fetching the balance. Still need to revert that change to address.test.tsx.

packages/components/src/components/Address/Address.test.tsx
Dhaiwat10: @hone1er no you're good haha

packages/components/src/components/TokenGate/TokenGate.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { ReactNode } from 'react';
+export interface TokenGateProps {
+  /**
+   * The provider or signer to fetch the address from the ens
+   */
+  walletBalance: number;
```
Dhaiwat10, line 6: You forgot to update the docstring! 😄

packages/components/src/components/TokenGate/TokenGate.tsx
```diff
@@ -0,0 +1,37 @@
...
+   */
+  deniedMessage?: ReactNode;
+}
+export const TokenGate: React.FC<TokenGateProps> = ({
+  walletBalance,
+  requiredQuantity = 1,
+  children,
+  deniedMessage,
+}) => {
+  console.log(walletBalance);
```
Dhaiwat10, line 26: Let's remove this `console.log`

packages/components/src/components/TokenGate/TokenGate.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  // return children within simple container(className optional)
```
Dhaiwat10, line 28:
```suggestion
  // return children within simple container
```

Dhaiwat10 (changes requested):
Looks good. Left some suggestions :)

hone1er (commented):
Implemented changes suggested by @Dhaiwat10

Fixed whitespace in Address.test.tsx to keep history clean. Removed console.log statement. Updated the docstring.

Dhaiwat10:
@allcontributors add @hone1er for code

PR #125, opened 2021-12-15
Add `switchToCorrectNetwork` method

### LINKED ISSUE #119: Switch to correct network
Whenever a user is on the wrong network, we will show a button to switch to the required network.
On clicking this button, the network will get changed to the required network instead of manually opening the MetaMask and changing the network.

Nazeeh21:
Closes #119 

## Description
This PR adds a method `switchToCorrectNetwork` to the `useWallet()` hook. This method allows the user to connect to the required network without manually heading over to the MetaMask and switching the network.

Users can use this method by directly assigning this `switchToCorrectNetwork` function to the `onClick` handler of any button and then the magic will happen and the MetaMask network will get changed to the required network.

In case, any user doesn't have MetaMask installed then it will show an alert to the user and ask them to install MetaMask.

### DIFF
--- a/packages/hooks/src/hooks/useWallet.ts
+++ b/packages/hooks/src/hooks/useWallet.ts
@@ -43,8 +43,10 @@ export function useWallet() {
           params: [{ chainId: `0x${requiredNetwork}` }], // chainId must be in hexadecimal numbers
         });
       } catch (error) {
-        // This error code indicates that the chain has not been added to MetaMask
-        // if it is not, then install it into the user MetaMask
+        console.error(error);
+        // This error code indicates that the chain has not been added to MetaMask.
+        // If it is not, then install it into the user MetaMask
+        // @ts-ignore
         if (error.code === 4902) {
           try {
             await window.ethereum.request({
@@ -60,7 +62,6 @@ export function useWallet() {
             console.error(addError);
           }
         }
-        console.error(error);
       }
     } else {
       // if no window.ethereum then MetaMask is not installed
--- a/packages/hooks/src/hooks/useWallet.ts
+++ b/packages/hooks/src/hooks/useWallet.ts
@@ -22,7 +22,7 @@ export function useWallet() {
     connected,
     provider,
     correctNetwork,
-    requiredNetwork,
+    network,
   } = context;
 
   React.useEffect(() => {
@@ -37,10 +37,11 @@ export function useWallet() {
     if (window.ethereum) {
       try {
         console.log('chainId', { chainId });
+        console.log('requiredNetwork', { network });
         // check if the chain to connect to is installed
         await window.ethereum.request({
           method: 'wallet_switchEthereumChain',
-          params: [{ chainId: `0x${requiredNetwork}` }], // chainId must be in hexadecimal numbers
+          params: [{ chainId: `0x${network}` }], // chainId must be in hexadecimal numbers
         });
       } catch (error) {
         console.error(error);
--- a/packages/hooks/src/hooks/useWallet.ts
+++ b/packages/hooks/src/hooks/useWallet.ts
@@ -46,8 +46,8 @@ export function useWallet() {
       } catch (error) {
         console.error(error);
         // This error code indicates that the chain has not been added to MetaMask.
-        // If it is not, then install it into the user MetaMask
-        // @ts-ignore
+        // If it is not, then install it into the user's MetaMask
+        // @ts-expect-error
         if (error.code === 4902) {
           try {
             await window.ethereum.request({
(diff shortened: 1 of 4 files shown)


Dhaiwat10:
@Nazeeh21 can you check the failing test?

Pponwili1998:
fix/issue-119

Pponwili1998:
maim

packages/hooks/src/stories/ConnectWallet.stories.tsx
```diff
@@ -1,16 +1,23 @@
-import React, { useEffect } from 'react';
 import { Button } from '@chakra-ui/react';
-import { NETWORKS, Provider, useWallet } from '..';
-import Authereum from 'authereum';
 import Portis from '@portis/web3';
+import Authereum from 'authereum';
+import React, { useEffect } from 'react';
+import { NETWORKS, Provider, useWallet } from '..';
```
Dhaiwat10, line 5: Why did you change the order of the imports? We should try to keep the changes to the minimum. Only introduce relevant changes so that it's easier for anyone to grasp the PR when re-visiting

packages/hooks/src/hooks/useWallet.ts
```diff
@@ -22,16 +22,56 @@ export function useWallet() {
...
+            });
+          } catch (addError) {
+            console.error(addError);
+          }
+        }
+      }
+    } else {
+      // if no window.ethereum then MetaMask is not installed
+      alert(
+        'MetaMask is not installed. Please consider installing it: https://metamask.io/download.html'
```
Dhaiwat10, line 70:
```suggestion
        'Switching networks automatically is only supported in MetaMask. https://metamask.io/download.html'
```

Can you change the error message?

packages/hooks/src/hooks/useWallet.ts
```diff
@@ -22,16 +22,56 @@ export function useWallet() {
...
+        // check if the chain to connect to is installed
+        await window.ethereum.request({
+          method: 'wallet_switchEthereumChain',
+          params: [{ chainId: `0x${network}` }], // chainId must be in hexadecimal numbers
+        });
+      } catch (error) {
+        console.error(error);
+        // This error code indicates that the chain has not been added to MetaMask.
+        // If it is not, then install it into the user MetaMask
+        // @ts-ignore
```
Dhaiwat10, line 50: Instead of `@ts-ignore`, did you try using something `@ts-expect-error`? I think that would be more appropriate.

packages/hooks/src/hooks/useWallet.ts
```diff
@@ -22,16 +22,56 @@ export function useWallet() {
...
+        console.log('requiredNetwork', { network });
+        // check if the chain to connect to is installed
+        await window.ethereum.request({
+          method: 'wallet_switchEthereumChain',
+          params: [{ chainId: `0x${network}` }], // chainId must be in hexadecimal numbers
+        });
+      } catch (error) {
+        console.error(error);
+        // This error code indicates that the chain has not been added to MetaMask.
+        // If it is not, then install it into the user MetaMask
```
Dhaiwat10, line 49:
```suggestion
        // If it is not, then install it into the user's MetaMask
```

packages/hooks/src/hooks/useWallet.ts
```diff
@@ -22,16 +22,56 @@ export function useWallet() {
...
+                },
+              ],
+            });
+          } catch (addError) {
+            console.error(addError);
+          }
+        }
+      }
+    } else {
+      // if no window.ethereum then MetaMask is not installed
```
Dhaiwat10, line 68:
```suggestion
      // if window.ethereum is undefined then MetaMask is not installed
```

Dhaiwat10 (changes requested):

Dhaiwat10:
You need to also run `yarn changeset` and do a minor bump to the `hooks` package

packages/hooks/src/stories/ConnectWallet.stories.tsx
Nazeeh21: There is not any specific reason for that. I'll change it back. 😅

packages/hooks/src/hooks/useWallet.ts
Nazeeh21: Sure, I'll update it.

packages/hooks/src/hooks/useWallet.ts
Nazeeh21: No, I haven't tried it, I'll try it.

Dhaiwat10 (approved):

PR #130, opened 2021-12-18
Fix NFT sizes and add docs

### LINKED ISSUE #94: Fix inconsistent NFT sizes
Looks like when we added gif and video support, it broke consistent sizing for the component:
image

I'll take on either resizing the content to fit in the box when it's a square, or adding padding around it so that we have consistency in sizing.

etr2460:
Closes #94 

## Description
Attempts to ensure all nft cards are squares of the same size. Also adds some more docs

## Screenshot
Screen Shot 2021-12-17 at 10 09 29 PM
Screen Shot 2021-12-17 at 10 09 34 PM
Screen Shot 2021-12-17 at 10 09 41 PM

to: @Dhaiwat10

### DIFF
--- a/packages/components/src/components/NFT/NFT.tsx
+++ b/packages/components/src/components/NFT/NFT.tsx
@@ -14,8 +14,18 @@ import {
 import fetch from 'cross-fetch';
 
 export interface NFTProps {
+  /**
+   * The address of the NFT contract.
+   */
   contractAddress: string;
+  /**
+   * The id of the NFT.
+   */
   tokenId: string;
+  /**
+   * The size of the NFT card.
+   */
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
 }
 
 export interface NFTData {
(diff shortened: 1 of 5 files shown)


packages/components/src/components/NFT/NFT.tsx
```diff
@@ -54,8 +64,12 @@ export const NFT = ({ contractAddress, tokenId }: NFTProps) => {
           animationUrl: data.animation_url,
         });
       }
-    } catch (error: any) {
-      setErrorMessage(error.message);
+    } catch (error) {
+      if (error instanceof Error) {
```
maximebonhomme, line 68: Could be good to create a helper for this as we're handling errors in quite a few places. What do you reckon? Can add a separate ticket for this.

etr2460: Makes sense. It seems like we're handling errors (and the typing of them) in a couple ways. I'll make an issue to figure out the best way to handle errors in our code, and use whatever helper function i make across the codebase

etr2460: https://github.com/Developer-DAO/web3-ui/issues/133

Dhaiwat10: @maximebonhomme feel free to create that ticket. Good suggestion

Dhaiwat10:
@etr2460 looks good. Can you add a changeset by running `yarn changeset`? I think we need a minor bump for the components package

etr2460:
Thanks for adding the changeset @Dhaiwat10. do you have docs around when to do a minor vs. a patch changeset? This PR was mainly a fix, although it did add a new (optional) prop to the NFT component

Dhaiwat10:
@etr2460 no I don't think we have any proper documentation around that. Do you think we need it?

PR #132, opened 2021-12-18
Address should not use input element

### LINKED ISSUE #124: Address Component: Should use Text instead of Input
### Description

We recently made some updates to the Address component and realised that it's rendering an `input` element instead of static text.

### Why is this needed?

This component is only used to display an address, it shouldn't need to use an `input`. Just some text would be enough.

### Proposed Solution or API

How about something like this:
```jsx
    <FormControl isInvalid={!!error}>
      <Flex
        alignItems='center'
        cursor={copiable ? 'pointer' : 'initial'}
        onClick={handleClick}
      >
        <Text>
          {displayAddress}
        </Text>
        {copiable && (
          <Box ml='auto'>
            {copied ? <CheckIcon color='green.500' /> : <CopyIcon color='gray.300' />}
          </Box>
        )}
      </Flex>
      <FormErrorMessage>{error}</FormErrorMessage>
    </FormControl>
```

This will display something unstyled - Up to the user to wrap it around a styled container.

maximebonhomme:
Closes #124 

## Description

Updated the Address component to use `Text` ChakraUI component instead of unnecessary `Input` component.

## 📝 Additional Information

It currently doesn't have any styling. It's up to the user to wrap it inside a styled container. I reckon let's wait until we have a design system before making a styled version of Address component?

### DIFF
--- a/packages/components/src/components/Address/Address.test.tsx
+++ b/packages/components/src/components/Address/Address.test.tsx
@@ -38,11 +38,11 @@ describe('Address copiable prop true', () => {
 
   it('uses writeText from Clipboard API', async () => {
     const { container } = render(<Address copiable value='taylorswift.eth' />);
-    const input = getByTestId(container, 'address-container');
+    const addressContainer = getByTestId(container, 'address-container');
 
     jest.spyOn(navigator.clipboard, 'writeText');
 
-    fireEvent.click(input);
+    fireEvent.click(addressContainer);
 
     await waitFor(() => {
       expect(navigator.clipboard.writeText).toHaveBeenCalledWith('taylorswift.eth');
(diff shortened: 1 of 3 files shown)


Dhaiwat10 (approved):
Thanks, looks good

packages/components/src/components/Address/Address.test.tsx
```diff
@@ -38,20 +38,19 @@ describe('Address copiable prop true', () => {
 
   it('uses writeText from Clipboard API', async () => {
     const { container } = render(<Address copiable value='taylorswift.eth' />);
-    const input = container.querySelector('input') as HTMLElement;
+    const addressContainer = getByTestId(container, 'address-container');
```
Dhaiwat10, line 41: Curious why did you go for `getByTestId` instead of `getByText`?

Dhaiwat10:
@allcontributors add @maximebonhomme for code

PR #134, opened 2021-12-20
Add `useTransaction` hook

### LINKED ISSUE #126: `useTransaction` hook
### Description

Yesterday, when trying to prepare a [Codesandbox demo](https://codesandbox.io/s/web3-ui-demo-ykvms) for `@web3-ui/hooks`, I felt that we could abstract out the `loading` and `error` state-handling for any transaction that goes through.

### Why is this needed?

The `error` & `loading` states can be better handled using a hook specifically designed to process one transaction.

### Proposed Solution or API

https://discord.com/channels/883478451850473483/920328787118419980/920328789353988148

https://gist.github.com/Dhaiwat10/7f1089acf4918c6b9ec202cd93b2b41c
https://www.diffchecker.com/5FK3pAa5

### Additional Information

_No response_

Nazeeh21:
Closes #126 

## Description
I've added a `useTransaction` hook. This hook takes the **function** and an array of **arguments** to this function as an argument. 
This hook returns an object containing the `execute` method, `loading` state, and `error` state. On calling this `execute` method, the function passed in the hook as an argument will get executed.

### DIFF
--- a/packages/hooks/src/stories/UseTransaction.stories.tsx
+++ b/packages/hooks/src/stories/UseTransaction.stories.tsx
@@ -1,37 +1,119 @@
-import { Button, Text } from '@chakra-ui/react';
+import { Button, Text, VStack, Input } from '@chakra-ui/react';
 import React from 'react';
-import { NETWORKS, Provider, useWallet } from '..';
+import { NETWORKS, Provider, useWallet, useContract } from '..';
 import { useTransaction } from '../hooks';
 
 export default {
   title: 'Hooks/useTransaction',
 };
 
-export const Default = () => {
-  const giveAlert = (name, greeting) => {
-    alert('Hello ' + name + '! ' + greeting);
-  };
+const ADDRESS = '0x7e1D33FcF1C6b6fd301e0B7305dD40E543CF7135'; // Rinkeby
+const ABI = [
+  {
+    inputs: [
+      {
+        internalType: 'string',
+        name: '_greeting',
+        type: 'string',
+      },
+    ],
+    stateMutability: 'nonpayable',
+    type: 'constructor',
+  },
+  {
+    inputs: [],
+    name: 'greet',
+    outputs: [
+      {
+        internalType: 'string',
+        name: '',
+        type: 'string',
+      },
+    ],
+    stateMutability: 'view',
+    type: 'function',
+  },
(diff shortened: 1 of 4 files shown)


packages/hooks/src/stories/UseTransaction.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { useTransaction } from '../hooks';
+
+export default {
+  title: 'Hooks/useTransaction',
+};
+
+export const Default = () => {
+  const giveAlert = (name, greeting) => {
+    alert('Hello ' + name + '! ' + greeting);
+  };
```
Dhaiwat10, line 13: This hook is meant to be for Ethereum contract calls. Can you add an example that actually calls an Ethereum contract? You can use `useContract`

Dhaiwat10: You can remove these current stories

Dhaiwat10 (changes requested):

packages/hooks/src/stories/UseTransaction.stories.tsx
Nazeeh21: > Sure, I'll use `useContract` and remove current stories

packages/hooks/src/stories/UseTransaction.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const { connectWallet, disconnectWallet, connected } = useWallet();
+  const contract = useContract(ADDRESS, ABI);
+  const [value, setValue] = React.useState('');
+
+  const [greet, greetLoading, greetError] = useTransaction(async () =>
+    // @ts-expect-error
+    alert(await contract.greet())
+  );
+
+  // @ts-expect-error
```
Dhaiwat10, line 74: Can you paste the error that you are seeing here?

Nazeeh21: I'm getting this typescript type error `Property 'setGreeting' does not exist on type '{}'.`

image

Dhaiwat10: That's expected since the `useContract` does not return any type defs as of now

PR #149, opened 2021-12-22
(chore) - remove extraneous deps

### LINKED ISSUE #128: Remove `@emotion/core` from `components` package's dependencies
I was trying to test out the combination of the `hooks` and `components` packages in a Codesandbox and I ran across this error after installing both of the packages. The error only appears if I also install the `hooks` package, not if I'm only using the `components` package.

image

[CodeSandbox link](https://codesandbox.io/s/sleepy-curran-u1rb3?file=/src/index.js)

I looked around a bit and came across this [StackOverflow comment](https://stackoverflow.com/a/65859704).

According to this comment, `@emotion/core` and `@emotion/react` are the same thing. We shouldn't have both of them in our dependencies.

### LINKED ISSUE #145: Remove `cross-fetch` from dependencies
Currently we're seeing a lot of unused dependencies and extraneous ones as well, currently we are leveraging `cross-fetch` in our components even though `effects` only run on the client. This means that for instance `fetchNFTData` and other relevant functions can just leverage the native browser fetch without us having to add the bundle-size for this. Most SSR frameworks inject their own fetch-polyfill in the runtime so they can facilitate caching, ....

JoviDeCroock:
Closes #145
Closes #128

## Description

This removes the extraneous `@emotion/core`, `cross-fetch` and `classnames` deps

### DIFF
--- a/packages/components/src/components/NFT/NFT.tsx
+++ b/packages/components/src/components/NFT/NFT.tsx
@@ -9,9 +9,8 @@ import {
   VStack,
   Skeleton,
   Alert,
-  AlertIcon,
+  AlertIcon
 } from '@chakra-ui/react';
-import fetch from 'cross-fetch';
 
 export interface NFTProps {
   /**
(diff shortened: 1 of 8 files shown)


JoviDeCroock:
Oh... We rely on actual fetching in our unit tests 😅

Dhaiwat10:
> Oh... We rely on actual fetching in our unit tests 😅

We have mocked the response though. Can we not do something about it? :(

JoviDeCroock:
yea fixed it with just polyfilling in the test-setup @Dhaiwat10

packages/components/src/components/NFT/NFT.tsx
```diff
@@ -47,7 +46,9 @@ export const NFT = ({ contractAddress, tokenId, size = 'xs' }: NFTProps) => {
 
   const fetchNFTData = useCallback(async () => {
     try {
-      const res = await fetch(`https://api.opensea.io/api/v1/asset/${contractAddress}/${tokenId}/`);
+      const res = await fetch(
```
JoviDeCroock, line 49: Most of these formatting changes are due to the lint-staged not actually applying correctly before it seems

PR #150, opened 2021-12-22
Spread ChakraProvider props to Provider Component

### LINKED ISSUE #137: Accept any props that a `ChakraProvider` accepts into our `components` `<Provider />`
The current `<Provider />` in the `components` package does not accept all the props that a `ChakraProvider` component. For example, if someone wanted to pass in a chakra theme to the `<Provider />` they wouldn't be able to. 

The [AddressInput](https://github.com/Developer-DAO/web3-ui/blob/main/packages/components/src/components/AddressInput/AddressInput.tsx#L34) component does this well.

Ibby-devv:
Closes #137

## Description

Spread built in ChakraProviderProps to Provider Component.

### DIFF
--- a/packages/components/src/components/Provider/Provider.tsx
+++ b/packages/components/src/components/Provider/Provider.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { ChakraProvider, ChakraProviderProps } from '@chakra-ui/react';
 
-export interface ProviderProps {}
+export interface ProviderProps extends ChakraProviderProps {}
 
-export const Provider: React.FC<ProviderProps & ChakraProviderProps> = ({ children, ...props }) => {
+export const Provider: React.FC<ProviderProps> = ({ children, ...props }) => {
   return <ChakraProvider {...props}>{children}</ChakraProvider>;
 };
--- a/packages/components/src/components/Provider/Provider.tsx
+++ b/packages/components/src/components/Provider/Provider.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
-import { ChakraProvider } from '@chakra-ui/react';
+import { ChakraProvider, ChakraProviderProps } from '@chakra-ui/react';
 
 export interface ProviderProps {}
 
-export const Provider: React.FC<ProviderProps> = ({ children }) => {
-  return <ChakraProvider>{children}</ChakraProvider>;
+export const Provider: React.FC<ProviderProps & ChakraProviderProps> = ({ children, ...props }) => {
+  return <ChakraProvider {...props}>{children}</ChakraProvider>;
 };
--- a/.changeset/tiny-bees-film.md
+++ b/.changeset/tiny-bees-film.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/components': minor
+---
+
+Accept any ChakraProvider-compatible props in our Provider


with-heart:
This makes me wonder if we should just get rid of the `Provider` component entirely until we have a reason for it. It's not doing anything except wrapping `ChakraProvider`.

packages/components/src/components/Provider/Provider.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
-import { ChakraProvider } from '@chakra-ui/react';
+import { ChakraProvider, ChakraProviderProps } from '@chakra-ui/react';
 
 export interface ProviderProps {}
```
with-heart, line 4: For some reason the GitHub suggestions interface is bugging out when I try to select multiple lines so I'll have to spread the suggestions across multiple comments 😬 

Since we're already using `interface` here, we can use `extends` to clean this up a bit.

```suggestion
export interface ProviderProps extends ChakraProviderProps {}
```

packages/components/src/components/Provider/Provider.tsx
```diff
@@ -1,8 +1,8 @@
...
 
-export const Provider: React.FC<ProviderProps> = ({ children }) => {
-  return <ChakraProvider>{children}</ChakraProvider>;
+export const Provider: React.FC<ProviderProps & ChakraProviderProps> = ({ children, ...props }) => {
```
with-heart, line 6: Then we don't need to use an intersection here since the type is already tied into `ProviderProps`.

```suggestion
export const Provider: React.FC<ProviderProps> = ({ children, ...props }) => {
```

with-heart (approved):

Ibby-devv:
> This makes me wonder if we should just get rid of the `Provider` component entirely until we have a reason for it. It's not doing anything except wrapping `ChakraProvider`.

Agree, not sure what the Provider is used for. If we need the ChakraProvider, why not just use it directly?

Dhaiwat10:
> This makes me wonder if we should just get rid of the Provider component entirely until we have a reason for it.

If we don't have a `Provider` that basically wraps everything inside the `ChakraProvider`, anyone wanting to use our library properly would have to install and set up Chakra UI themselves. I don't think we would want that.

Dhaiwat10:
> If we need the ChakraProvider, why not just use it directly?

That's what we are doing! We're just indirectly shipping it as our own provider. We're assuming that not everyone using this package needs to be educated about Chakra.

Ibby-devv:
> > If we need the ChakraProvider, why not just use it directly?
> 
> That's what we are doing! We're just indirectly shipping it as our own provider. We're assuming that not everyone using this package needs to be educated about Chakra.

What I meant was our Provider could simply use the ChakraProvider directly until we have something more to add to it.

Dhaiwat10:
@Ibby-devv that's precisely what it's doing right now, I think. I am aware that we could just directly export the ChakraProvider as our own Provider but I prefer having a separate file and declaration like this. Purely my opinion though.

Ibby-devv:
> @Ibby-devv that's precisely what it's doing right now, I think. I am aware that we could just directly export the ChakraProvider as our own Provider but I prefer having a separate file and declaration like this. Purely my opinion though. 

Yeah that's fair enough. I'm on my phone so not sure if the changes I made were what you were looking for in terms of passing on the props to our component.

Dhaiwat10:
I merged them in, they're fine!

PR #152, opened 2021-12-22
chore: clean up build

### LINKED ISSUE #136: Build step failing with `TS2742` error
PRs with changes unrelated to `useWallet` appear to be failing at the Build step with the following message:

```
🎁 error src/hooks/useWallet.ts:8:17 - error TS2742: The inferred type of 'useWallet' cannot be named without a reference to 'ethers/node_modules/@ethersproject/providers'. This is likely not portable. A type annotation is necessary.
```

Example failure: https://github.com/Developer-DAO/web3-ui/runs/4571409517?check_suite_focus=true#step:7:15

with-heart:
This PR makes a number of changes to the repo in order to consolidate and clean up the build process. There's a lot of unnecessary stuff in the repo that doesn't need to exist in order for `preconstruct` to build the packages, so I removed a lot of it.

Here's a list:

- replaced individual package `.babelrc` files with a root `babel.config.js` file
- removed unused `example/package-lock.json` file
- consolidated the various `build` scripts into a single root-level `build` script
- switched to `dependencies` instead of `devDependencies` (not a required change by any means, but in the root monorepo `package.json`, there's no distinction between deps and devDevs)
- removed some `babel` dependencies that aren't required
- removed the `outDir` config from `tsconfig.json` which was breaking builds after all of the other changes

This was originally intended to close #136 but it appears that was already fixed by something else, so this PR is just icing on the cake

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,11 @@
+const BABEL_ENV = process.env.BABEL_ENV;
+const cjs = BABEL_ENV === 'cjs';
+
+module.exports = {
+  presets: [
+    '@babel/preset-env',
+    '@babel/preset-typescript',
+    '@babel/preset-react',
+  ],
+  plugins: ['@babel/plugin-transform-runtime'],
+};
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -37,6 +37,7 @@
   "module": "dist/web3-ui-components.esm.js",
   "types": "dist/web3-ui-components.cjs.d.ts",
   "dependencies": {
+    "@babel/runtime": "^7.16.7",
     "@chakra-ui/icons": "^1.1.1",
     "@chakra-ui/react": "^1.7.2",
     "@emotion/react": "^11",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -29,7 +29,6 @@
     "dist"
   ],
   "scripts": {
-    "build": "preconstruct build",
     "format": "prettier --write \"src/**/*.{ts,tsx,json,js,jsx}\"",
     "format:check": "prettier --list-different \"src/**/*.{ts,tsx,json,js,jsx}\"",
     "prepublishOnly": "yarn build && cp ../../LICENSE ./LICENSE"
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -37,7 +37,6 @@
   "module": "dist/web3-ui-components.esm.js",
   "types": "dist/web3-ui-components.cjs.d.ts",
   "dependencies": {
-    "@babel/runtime": "^7.16.3",
     "@chakra-ui/icons": "^1.1.1",
     "@chakra-ui/react": "^1.7.2",
     "@emotion/react": "^11",
@@ -50,7 +49,6 @@
     "react-dom": "*"
   },
   "devDependencies": {
-    "@babel/core": "^7.12.7",
     "@storybook/react": "^6.3.12",
     "@types/jest": "^26.0.15",
     "@types/node": "^16.11.9",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,22 +2,6 @@
 # yarn lockfile v1
 
 
-"@babel/cli@^7.16.0":
-  version "7.16.0"
-  resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.16.0.tgz#a729b7a48eb80b49f48a339529fc4129fd7bcef3"
-  integrity sha512-WLrM42vKX/4atIoQB+eb0ovUof53UUvecb4qGjU2PDDWRiZr50ZpiV8NpcLo7iSxeGYrRG0Mqembsa+UrTAV6Q==
-  dependencies:
-    commander "^4.0.1"
-    convert-source-map "^1.1.0"
-    fs-readdir-recursive "^1.1.0"
-    glob "^7.0.0"
-    make-dir "^2.1.0"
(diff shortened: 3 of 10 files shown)


packages/components/package.json
```diff
@@ -27,7 +26,6 @@
   "module": "dist/web3-ui-components.esm.js",
   "types": "dist/web3-ui-components.cjs.d.ts",
   "dependencies": {
-    "@babel/runtime": "^7.16.3",
```
JoviDeCroock, line 30: without `@babel/runtime` the regenerator runtime won't work https://www.runpkg.com/?@web3-ui/hooks@0.5.0/dist/web3-ui-hooks.esm.js#1 the bundles aren't too ideal in terms of build composition 😅

yarn.lock
```diff
@@ -1115,13 +1099,20 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
-"@babel/runtime@^7.0.0", "@babel/runtime@^7.10.2", "@babel/runtime@^7.10.4", "@babel/runtime@^7.12.13", "@babel/runtime@^7.12.5", "@babel/runtime@^7.13.10", "@babel/runtime@^7.14.0", "@babel/runtime@^7.14.8", "@babel/runtime@^7.16.3", "@babel/runtime@^7.3.1", "@babel/runtime@^7.5.0", "@babel/runtime@^7.5.5", "@babel/runtime@^7.7.2", "@babel/runtime@^7.7.6", "@babel/runtime@^7.7.7", "@babel/runtime@^7.8.4", "@babel/runtime@^7.9.2":
+"@babel/runtime@^7.0.0", "@babel/runtime@^7.10.2", "@babel/runtime@^7.12.13", "@babel/runtime@^7.12.5", "@babel/runtime@^7.13.10", "@babel/runtime@^7.14.0", "@babel/runtime@^7.14.8", "@babel/runtime@^7.16.3", "@babel/runtime@^7.3.1", "@babel/runtime@^7.5.0", "@babel/runtime@^7.7.2", "@babel/runtime@^7.7.6", "@babel/runtime@^7.7.7", "@babel/runtime@^7.8.4", "@babel/runtime@^7.9.2":
```
JoviDeCroock, line 1102: seeing these makes me think we need `yarn-deduplicate` in the `pre-commit` when there are `yarn.lock` changes

babel.config.js
```diff
@@ -0,0 +1,6 @@
+const BABEL_ENV = process.env.BABEL_ENV;
+const cjs = BABEL_ENV === 'cjs';
+
+module.exports = {
+  presets: ['@babel/preset-env', '@babel/preset-typescript', '@babel/preset-react'],
```
JoviDeCroock, line 5: Due to a bug (?) in preconstruct we need `plugins: ["@babel/plugin-transform-runtime"]` here else the `async await` syntax will be erroneous in the eventual output build

packages/components/package.json
with-heart: I see the helpers inlined in the build output locally?

yarn.lock
with-heart: It's weird because I don't even see half of the versions listed here as output from `yarn why @babel/runtime`

packages/components/package.json
JoviDeCroock: oh weird, this used to bug out before

babel.config.js
with-heart: Hmmmm, that's interesting. I don't think any of the `preconstruct` projects I work on use this and I've not noticed any issues with `async`/`await`

packages/components/package.json
with-heart: I see what you're saying locally. That… doesn't make any sense.

JoviDeCroock: Yes, it caught me off-guard as well when I noticed 😅 judging from their issues it feels "intended" but then again I always try to avoid async await in my OSS projects so not too familiar with the matter

with-heart: I wonder what would cause this. This is similar to the other comment in that I've never had to use this before in any other projects and haven't run into issues. 

I see a few things referencing runtime-related things in this issue: https://github.com/preconstruct/preconstruct/issues/297#issuecomment-690964802

But there's nothing that jumps out to me that this should be required. I feel like we should open a new issue over there but I'm not even sure what to report exactly 😆

with-heart:
I'll roll back the `@babel/runtime` change since there's some weirdness around it

Dhaiwat10:
Thanks for this @with-heart ❤️  Can I merge this?

Dhaiwat10:
Sorry for pinging you again @with-heart but is this ready to be merged? 😄

JoviDeCroock:
@Dhaiwat10 in current state this breaks the build as mentioned in the conversations here

Dhaiwat10:
@JoviDeCroock that's strange because according to Github all checks are passing. Maybe we need to re-run the workflow?

JoviDeCroock:
@Dhaiwat10 a workflow runs against the source code, the build is broken in the eventual output as per https://github.com/Developer-DAO/web3-ui/pull/152#discussion_r774202453

Dhaiwat10:
I see, thanks

with-heart:
I'll get this caught up and ready to merge here soon :)

Dhaiwat10:
I know nothing about this topic so I'll let @JoviDeCroock review. I'll watch and learn :3

JoviDeCroock (approved):
Looks good! Awesome work

PR #155, opened 2021-12-24
Fix/network change hooks

### LINKED ISSUE #153: UseWallet hook throws error if wallet is disconnected through metamask
After connecting with the useWallet hook, if the user disconnects directly from the wallet 2 errors are thrown: "missing provider" and "unknown account #0(operation="getAddress"....)"

See video below

https://user-images.githubusercontent.com/24376928/147302858-9a6c877f-da33-4f7b-b5a0-bdb0d71ad9bd.mov

The code being used. This is a basic next-js app
Screen Shot 2021-12-23 at 4 52 35 PM
Screen Shot 2021-12-23 at 4 52 21 PM

### LINKED ISSUE #154: [Hooks] `connection` is `null` when the network changes
The value of `connection` is coming out as `null` when `onNetworkChange` is called. This is causing a crash whenever someone tries to change the network once connected. 

https://github.com/Developer-DAO/web3-ui/blob/main/packages/hooks/src/Provider.tsx#L119

Dhaiwat10:
Closes #153 
Closes #154 

- Refactored the `connectWallet` function because the value of `connection` wasn't being persisted. We were using the `connection` object as our 'provider' and hence it was throwing errors in methods like `onNetworkChanage`.

- Also added an edge case inside `accountsChanged` to detect if no accounts are connected anymore.

### DIFF
--- a/.changeset/lucky-starfishes-bow.md
+++ b/.changeset/lucky-starfishes-bow.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/hooks': patch
+---
+
+Fix issues related to accountChange and networkChange
--- a/packages/hooks/src/Provider.tsx
+++ b/packages/hooks/src/Provider.tsx
@@ -16,7 +16,9 @@ export interface Web3ContextType {
   network: number;
 }
 
-export const Web3Context = React.createContext<Web3ContextType | undefined>(undefined);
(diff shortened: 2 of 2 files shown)


Dhaiwat10:
There are also some unrelated formatting changes because of the recently updated prettier config.

PR #110, opened 2021-12-13
avatar-component

jwarshack:
Closes #110 

## Description

> Add a brief description
Created a component for ENS avatars using @react/davatar library.

## 📝 Additional Information

Component takes in an address and displays ENS avatar if applicable. If not, displays a jazzicon. Size is customizable but default value is 50px.

Screen Shot 2021-12-13 at 9 16 02 AM

PR #157, opened 2021-12-24
Avatar component

### LINKED ISSUE #66: `<Avatar />`
### Description

An Avatar component that renders the avatar NFT linked with an ENS would be handy. 

https://defireport.io/nft/ens-now-supports-nft-avatars-for-your-profile/

### Why is this needed?

Quite a few people are using their ENS as their online identity now which includes linking one of their NFTs to their ENS as their avatar. It would be great if developers had an handy component that renders this avatar.

### Proposed Solution or API

```tsx
<Avatar address='dhaiwat.eth' />

// or

<Avatar address='0x0ED6Cec17F860fb54E21D154b49DAEFd9Ca04106' />
```

Dhaiwat10:
Originally #110
Closes #66

## Description

> Add a brief description
Created a component for ENS avatars using @react/davatar library.

## 📝 Additional Information

Component takes in an address and displays ENS avatar if applicable. If not, displays a jazzicon. Size is customizable but default value is 50px.

Screen Shot 2021-12-13 at 9 16 02 AM


packages/components/src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Button } from '@chakra-ui/react';
+
+export default {
+  title: 'Components/Avatar',
+  component: Avatar,
+};
+
+export const Default = () => (
+  <Avatar address='0x983110309620D911731Ac0932219af06091b6744' size={50} />
+);
```
etr2460, line 16: can we add a story for when the address doesn't have an avatar?

packages/components/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export interface AvatarProps {
+  /**
+   * The resolved address
+   */
+  address: string;
+  /**
+   * The size of the avatar
+   */
+  size?: number;
```
etr2460, line 12: should we try to use Chakra-UI sizes here? so that this accepts `xs`, `sm`, etc. too?

etr2460 (approved):
a few comments, nothing blocking, but would be nice to have

with-heart:
I think we should consider naming this something other than `Avatar`. Since we're based on Chakra which already exports an `Avatar` component, we're setting our users up for a potentially confusing situation. I'd suggest `EnsAvatar` since it's a specific type of avatar.

with-heart (changes requested):
After looking at the source, I'd suggest that we close this. If all we're doing with the component is re-exporting another library's component (and making it less flexible in the process, since we reduce the props down to `size` and `avatar` only), I think we should recommend that users check out the other lib rather than including it directly in our lib.

If we built something on top of that library unique to `web3-ui`, it'd make a lot more sense to include that here.

Dhaiwat10:
@with-heart that definitely sounds sensible and something I can agree with. What are your thoughts on this? @jwarshack @etr2460

jwarshack:
@Dhaiwat10 @etr2460 Agree with @with-heart. The only benefit would be convenience but probably not worth the tradeoff of less flexibility.

Dhaiwat10:
@jwarshack Let's close this then! Sorry that your PR is going down the drain. We def need to learn from this and make sure we think things over before working on something. Thanks Jake ❤️

Dhaiwat10:
@allcontributors please add @jwarshack for code

PR #162, opened 2021-12-24
Add null check to address component

### LINKED ISSUE #151: Address component throws error when shortened=true
Address component throws error when shortened=true.

Screen Shot 2021-12-22 at 12 07 01 PM
It works correctly when i set shortened to the line below but maybe it can be solved at the component level so the dev can just set true or false without getting this error.
Screen Shot 2021-12-22 at 12 06 40 PM
 So i'm assuming the Address component needs to wait for the userAddress before it tries to shorten it.

But this could also be user error on my part for not checking the connection before passing the value since the below code works as well. In which case we can close this issue
Screen Shot 2021-12-22 at 12 20 59 PM

hone1er:
Closes #151 

## Description

Added a null check for value prop to resolve "cannot read properties of null" error


Dhaiwat10:
@hone1er yo I think the history for this branch is a little messed up. I tried rebasing it but ran into too many conflicts. Can you try starting a branch afresh from the latest main branch and put in a new PR so that it has all the latest code?

Dhaiwat10:
Oh, ok nevermind I think we can get away with this by squash-merging this PR

.changeset/two-turtles-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@web3-ui/components': minor
+---
+
+Added a TokenGate component that restricts access to child components unless erc-20/erc-721 token quantity requirements are met.
```
Dhaiwat10, line 5: Or not...we can't merge this one 😅  Create a new branch afresh pls!

hone1er: Haha alright for sure will do

PR #164, opened 2021-12-24
added null check for value prop

### LINKED ISSUE #151: Address component throws error when shortened=true

hone1er:
Closes #151 
## Description

Added a null check for the "value" prop to resolve "Cannot read properties of null" error.

### DIFF
--- a/.changeset/yellow-eyes-eat.md
+++ b/.changeset/yellow-eyes-eat.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/components': minor
+---
+
+Bug fix. Added null check to address component for value prop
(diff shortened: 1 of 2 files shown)


Dhaiwat10:
Sweet

PR #168, opened 2021-12-25
usePoller Hook

Ibby-devv:
A hook that allows the calling of a function at dedicated intervals.

## Description

A simple hook that allows calling a function at set intervals. Useful for updating the UI of your Dapp.

## Usage

usePoller (callbackFn, delay)

### DIFF
--- a/packages/hooks/src/hooks/usePoller.ts
+++ b/packages/hooks/src/hooks/usePoller.ts
@@ -1,6 +1,18 @@
 import { useEffect, useRef } from "react";
 
 export const usePoller = (func: () => void, delay: number): void => {
+ 
+  /**
+ * Calls a function at a set interval. The function will also be called
+ * immediately.
+ *
+ * @param func The function to call at an interval
+ * @param delay The delay between each interval call
+ * @example
+ * const callback = () => console.log('test')
+ * usePoller(callback, 1000) // logs 'test' every second
+  */
+  
   const savedCbFunc = useRef(func);
 
   // Remember the latest fn.
--- a/packages/hooks/src/hooks/usePoller.ts
+++ b/packages/hooks/src/hooks/usePoller.ts
@@ -1,22 +1,20 @@
 import { useEffect, useRef } from "react";
 
 const usePoller = (func: () => void, delay: number): void => {
-  const savedCbFunc = useRef<() => void>();
+  const savedCbFunc = useRef(func);
 
   // Remember the latest fn.
-  useEffect((): void => {
+  useEffect(() => {
     savedCbFunc.current = func;
   }, [func]);
 
-  useEffect((): void | (() => void) => {
-    function tick() {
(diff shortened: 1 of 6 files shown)


packages/hooks/src/hooks/usePoller.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useRef } from "react";
+
+const usePoller = (func: () => void, delay: number): void => {
```
Dhaiwat10, line 3: Are we sure that the functions we pass to this hook will always return `void`? Most web3 contract calls return a `Promise`

Ibby-devv: I typically wrap any code I use with this hook into a function that returns void. You could make it super generic and make it of type any if you wanted to.

with-heart, line 3: We should attach some documentation to this function using `jsdoc` so users can understand its purpose:

```suggestion
/**
 * Calls a function at a set interval. The function will also be called
 * immediately.
 *
 * @param func The function to call at an interval
 * @param delay The delay between each interval call
 * @example
 * const callback = () => console.log('test')
 * usePoller(callback, 1000) // logs 'test' every second
 */
const usePoller = (func: () => void, delay: number): void => {
```

with-heart, line 3: We should follow the pattern for the other hooks and export this directly rather than using a `default` export.

```suggestion
export const usePoller = (func: () => void, delay: number): void => {
```

packages/hooks/src/hooks/usePoller.ts
```diff
@@ -0,0 +1,27 @@
...
+      return () => clearInterval(id);
+    }
+  }, [delay]);
+
+  useEffect(() => {
+    func();
+  }, []);
+};
+
+export default usePoller;
```
with-heart, line 27:
```suggestion
```

with-heart (changes requested):

Dhaiwat10:
Thanks for the review @with-heart Really appreciate a helping hand, can't review everything myself <3

packages/hooks/src/hooks/usePoller.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    if (delay !== null) {
+      const id = setInterval(tick, delay);
+      return () => clearInterval(id);
+    }
+  }, [delay]);
+
+  useEffect(() => {
+    func();
+  }, []);
```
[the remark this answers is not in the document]
Dhaiwat10: @Ibby-devv can you look into this please? :) I think we're good to merge once this is addressed.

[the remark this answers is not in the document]
Ibby-devv: Missed this one, should be there now

Dhaiwat10:
@allcontributors please add @Ibby-devv for code!

PR #180, opened 2021-12-26
Create test for useWallet

### LINKED ISSUE #176: Add tests for `useWallet`

gjsyme:
Closes #176 

## Description

Add @testing-library/react-hooks, make use of Provider for context and test that useWallet can instantiate and trigger the request to connect to the correct network.

## 📝 Additional Information

Pretty basic tests, but extensible if we want to either deeply mock window.ethereum, make use of another external library (though none jump out as best-choice), or solve this in another way.

### DIFF
--- a/packages/hooks/src/hooks/useWallet.test.tsx
+++ b/packages/hooks/src/hooks/useWallet.test.tsx
@@ -13,42 +13,47 @@ afterEach(() => {
   windowSpy.mockRestore();
 });
 
-test('should use the wallet', () => {
-  const wrapper = ({ children }) => (
-    <Provider network={1} infuraId={'abc123'}>
-      {children}
-    </Provider>
-  );
-  const { result } = renderHook(() => useWallet(), { wrapper });
+describe('useWallet tests', () => {
+  test('should use the wallet', () => {
+    const wrapper = ({ children }) => (
+      <Provider network={1} infuraId={'abc123'}>
+        {children}
+      </Provider>
+    );
+    const { result } = renderHook(() => useWallet(), { wrapper });
 
-  expect(result.current.connected).toBe(false);
-  expect(typeof result.current.connectWallet).toBe('function');
-});
+    expect(result.current.connected).toBe(false);
+    expect(typeof result.current.connectWallet).toBe('function');
+  });
 
-// initially is chainId undefined / requires network per the argument
-test('should be able to switch networks', () => {
-  const testNetwork = 2;
-  const wrapper = ({ children }) => (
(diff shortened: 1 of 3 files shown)


etr2460:
The chromatic run failed because this PR is made from a fork. If you want to, then you could re open the PR from a branch in the repo. that said, since it's only a test, i feel like it would be safe without chromatic running

packages/hooks/src/hooks/useWallet.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+let windowSpy;
+
+beforeEach(() => {
+  windowSpy = jest.spyOn(window, 'window', 'get');
+});
+
+afterEach(() => {
+  windowSpy.mockRestore();
+});
```
etr2460, line 14: i'm pretty sure these should go inside the describe block

packages/hooks/src/hooks/useWallet.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+describe('useWallet tests', () => {
+  test('should use the wallet', () => {
+    const wrapper = ({ children }) => (
+      <Provider network={1} infuraId={'abc123'}>
```
etr2460, line 19: this is typically how react passes strings as props

```suggestion
      <Provider network={1} infuraId="abc123">
```

packages/hooks/src/hooks/useWallet.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    expect(result.current.connected).toBe(false);
+    expect(typeof result.current.connectWallet).toBe('function');
+  });
+
+  // initially is chainId undefined / requires network per the argument
+  test('should be able to switch networks', () => {
+    const testNetwork = 2;
+    const wrapper = ({ children }) => (
+      <Provider network={testNetwork} infuraId={'abc123'}>
```
etr2460, line 33:
```suggestion
      <Provider network={testNetwork} infuraId="abc123">
```

etr2460 (commented):
a few comments but mostly good. thanks for adding tests!

packages/hooks/src/hooks/useWallet.test.tsx
gjsyme: agreed. been using something that has gotten me in the habit of objectifying my strings and i shouldn't.

etr2460 (approved):
let's do it, thanks for the tests!

etr2460:
@allcontributors add @gjsyme for code

Issue #182, opened 2021-12-27
Setup a Next.js example


Dhaiwat10:
At the moment, we don't have an example app demonstrating our project. We should have one. (The `example` folder in the main branch is broken and doesn't do anything)

I tried creating a Next.js app inside the `example` folder at the root of the project using `yarn create next-app example --ts` and ended up with the usual Next.js starter template. I then did `yarn add @web3-ui/core` inside the `example` folder. This broke the app. (see SS below)
![image](https://user-images.githubusercontent.com/39617427/147489956-e51ef8df-e38e-44cc-af32-b1b7673939af.png)

Can someone help me fix this so that we can import stuff from our packages and this example add doesn't break?

Ibby-devv:
Did you add Web3ui/core as a dependancy?

PR #183, opened 2021-12-29
Setup nextjs demo page

etr2460:
Closes https://github.com/Developer-DAO/web3-ui/issues/182

## Description
Fixes a few issues with the next js demo setup and adds a very basic demo page that verifies components work. Also converts everything to using TS

## 📝 Additional Information
There's probably still more work to be done to make a good demo page. I'm also a bit uncertain if a demo page is necessary in addition to the storybook. Food for thought

## Screenshot
image


with-heart:
> I'm also a bit uncertain if a demo page is necessary in addition to the storybook. Food for thought

Yeah, I think you're right to be uncertain there. imo the example should become a real-world app, something that gives us a way to demonstrate realistic use cases for each piece of `web3-ui` (and show off our creativity too! 😁).

with-heart (changes requested):
Can you remove `example/package-lock.json` and make sure it's working as a workspace with `yarn`?

Dhaiwat10:
> imo the example should become a real-world app, something that gives us a way to demonstrate realistic use cases for each piece of web3-ui

That's precisely what I was thinking. The Storybook is great at showcasing the components and hooks in isolation, but something that people can play around with directly (we will deploy this example to Vercel or sth) and maybe clone the code and try it out for themselves would be nice too. Maybe this can even turn into a starter template with our package all setup and ready-to-go, `npx create-web3-ui-app`, who knows?

example/pages/_app.tsx
```diff
@@ -3,7 +3,7 @@ import { Provider } from '@web3-ui/core';
 
 function MyApp({ Component, pageProps }) {
   return (
-    <Provider network={4}>
+    <Provider network={1}>
```
Dhaiwat10, line 6:
```suggestion
    <Provider network={NETWORKS.mainnet}>
```

More readable. `NETWORKS` is exported from `core` and `hooks`!

Dhaiwat10 (approved):
Just one tiny change. I'll do this myself, push and merge :)

Dhaiwat10:
Oh I missed @with-heart's review. Will also try to have a look at that.

Dhaiwat10:
image

I can't run this locally. I'm running `yarn example` from root. Anyone else facing the same problem?

etr2460:
>Can you remove example/package-lock.json and make sure it's working as a workspace with yarn?

Cross posting from discord, the examples directory was initially setup with npm, not yarn. i'm not sure if that was intentional. Maybe there's some questions here to be answered first:
- Should we have an example site, or is the storybook sufficient?
- Should the example use npm or yarn?

Once we answer both of these, we can decide if this PR should be updated, or if we should simply delete the examples directory for now

Dhaiwat10:
> Should we have an example site, or is the storybook sufficient?

I think we do need an example because of the reasons I mentioned [here](https://github.com/Developer-DAO/web3-ui/pull/183#issuecomment-1002510328). Tldr; storybook is good for showcasing stuff in isolation but we need an real-world 'app' example where people can see how all of that stuff actually works together.

> Should the example use npm or yarn?

It should use yarn because that's what we are using everywhere else in the project. The package-lock.json was unintentional. Sorry for that.

example/pages/_app.tsx
etr2460: fixed

etr2460:
@Dhaiwat10, @with-heart this should be g2g now

package.json
```diff
@@ -24,7 +24,7 @@
     "test:coverage": "jest --coverage --colors",
     "storybook": "start-storybook -p 9001 -s ./packages/components/src/assets -c .storybook",
     "postinstall": "preconstruct dev",
-    "example": "yarn workspace example dev",
+    "example": "yarn build && yarn workspace example dev",
```
etr2460, line 27: for nextjs to load the packages, we need to `yarn build` them first. This isn't technically correct as the examples should depend on a version of the packages and not the `main` branch of them, but I couldn't for the life of me figure out how to get yarn to pull in the packages from npm instead of symlinks. Maybe there was a caching issue, but this seems good enough for me

Dhaiwat10:
Way to go @etr2460 ❤️

Dhaiwat10 (approved):

PR #185, opened 2021-12-29
Create a status indicator for useContract

### LINKED ISSUE #167: Loading status indicator / 'readiness' indicator for `useContract`
Contracts take some time to load. Right now, if someone wrote this:

```tsx
const greeterContract = useContract(xxx,xxx);
```

Then `greeterContract` will go through the following lifecycle:

```
null          ---->           { } (empty object)        ---->            actual contract object (ready)
```

Because of that weird empty object state in the middle, just doing
```ts
if (!greeterContract)
```
doesn't work.

We need to ship an indicator that notifies the user when the contract is ready. Something like:

```tsx
const [contract, ready] = useContract(___, ___);

// ready becomes true when the contract has finished loading
```

CatalinCsnMaster:
Closes #167 

## Description

Create status indicator `isReady` for `useContract` hook.

### DIFF
--- a/.changeset/short-eggs-juggle.md
+++ b/.changeset/short-eggs-juggle.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/hooks': minor
+---
+
+useContract status indicator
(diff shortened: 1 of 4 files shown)


Dhaiwat10:
@allcontributors add @CatalinCsnMaster for code ❤️

PR #189, opened 2021-12-29
Add check for `context.connected` in `useContract` for readiness

Dhaiwat10:
This PR introduces a new check in the `useContract` hook which only attempts to initalize the contract if the user has connected their wallet. Calling contracts when the user hadn't connected their wallet was throwing errors.

### DIFF
--- a/.changeset/rare-days-punch.md
+++ b/.changeset/rare-days-punch.md
@@ -2,4 +2,4 @@
 '@web3-ui/hooks': patch
 ---
 
-Add check for user connection during contract init in useContract
+`useContract` now initializes the contract only after the user has connected their wallet. This helps prevent errors from calling contracts without a connection.
--- a/.changeset/rare-days-punch.md
+++ b/.changeset/rare-days-punch.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/hooks': patch
+---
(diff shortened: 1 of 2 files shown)


packages/hooks/src/hooks/useContract.ts
```diff
@@ -7,7 +7,7 @@ export function useContract(address: string, abi) {
   const [contract, setContract] = React.useState({});
   const [isReady, setIsReady] = React.useState(false);
   React.useEffect(() => {
-    if (context) {
+    if (context && context.connected) {
```
with-heart, line 10:
```suggestion
    if (context?.connected) {
```

with-heart (approved):

.changeset/rare-days-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@web3-ui/hooks': patch
+---
+
+Add check for user connection during contract init in useContract
```
with-heart, line 5: I think this would probably give a little extra context:

```suggestion
`useContract` now initializes the contract only after the user has connected their wallet. This helps prevent errors from calling contracts without a connection.
```

with-heart (approved):

PR #198, opened 2021-12-31
Add generic type argument for useContract

### LINKED ISSUE #197: Return contract function type definitions from `useContract`
At the moment, the contract instance that the `useContract` hook returns is just a simple contract object returned from an `ethers` contract initialization. It does not contain any type definitions. We should add this.

```tsx
const [greeterContract, ready] = useContract(ADDRESS, ABI);

console.log(typeof greeterContract.greeting); // should be an async function that returns a string

console.log(typeof greeterContract.setGreeting); // should be an async function that accepts a string
```

carlomigueldy:
Closes #197 

## Description

Allows to auto-generate TypeScript smart contract type definitions based on the ABIs provided by the user.

Though it can't be generated in runtime but will have to let user do additional step to generate these typings. We can explore some other ideas if anyone has a better approach to this, the way I see how the developer would be generating types is to execute a command from `typechain`, which should be additional dependency for their side... 

1. Given that they only installed this hooks package, and `useContract` is already being imported and used
2. Install `typechain` library by `yarn add typechain --dev` or `npm i -D typechain`
3. Add a script into their `package.json` file, the script can be named anything they prefer but recommend "typechain"
```json
"scripts": {
    "typechain": "typechain --target=ethers-v5 ABI_DIRECTORY --out-dir=OUTPUT_DIRECTORY",
}
```
- The ABI_DIRECTORY will be containing all of the ABIs that they have saved, so as an example they chose to store all ABI in this directory `src/abis/` but it should be defined as `src/abis/**/*.json` in the script. i.e. `typechain --target=ethers-v5 src/abis/**/*.json --out-dir=OUTPUT_DIRECTORY`
- The OUTPUT_DIRECTORY will be the preferred output directory for the developer and it is recommended that it should be on `.gitignore` since these are generated files and it can be generated once again when required, so it's fine to have this ignored. As an example the output directory can be defined as `src/types/contracts` in the script. i.e. `typechain --target=ethers-v5 src/abis/**/*.json --out-dir=src/types/contracts`
```json
"scripts": {
    "typechain": "typechain --target=ethers-v5 src/abis/**/*.json --out-dir=src/types/contracts",
}
```
4. As soon as they have the "typechain" script added, and the ABIs are in place in the directory (i.e. ABI_DIRECTORY) they store it, then finally execute the script to generate these typings
```bash
yarn typechain

# or 

npm run typechain
```

## 📝 Additional Information

I also sort of removed some code from the `useContract` since I just thought it was not necessary anymore given that we'll be utilizing Typechain for auto-generating type definitions on the fly.

**PS:** And at the time I am writing this I still haven't tested it out myself on an example application. This will be on DRAFT. When I tested it out and my assumption was correct on the approach then I will be removing this.

### DIFF
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -29,16 +29,13 @@ function MyApp({ Component, pageProps }) {
 
 The following hooks are available:
 
-- [@web3-ui/hooks](#web3-uihooks)
-  - [Getting started](#getting-started)
-  - [Hooks](#hooks)
-    - [useWallet](#usewallet)
-    - [useContract](#usecontract)
-    - [useTransaction](#usetransaction)
-    - [useTokenBalance](#usetokenbalance)
-    - [useReadOnlyContract](#usereadonlycontract)
-    - [useReadOnlyProvider](#usereadonlyprovider)
-    - [usePoller](#usepoller)
+- [useWallet](#usewallet)
+- [useContract](#usecontract)
+- [useTransaction](#usetransaction)
+- [useTokenBalance](#usetokenbalance)
+- [useReadOnlyContract](#usereadonlycontract)
+- [useReadOnlyProvider](#usereadonlyprovider)
+- [usePoller](#usepoller)
 
 ---
 
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -29,10 +29,13 @@ function MyApp({ Component, pageProps }) {
 
 The following hooks are available:
 
-- [useWallet](#usewallet)
-- [useContract](#usecontract)
-- [useTransaction](#usetransaction)
-- [useTokenBalance](#usetokenbalance)
+- [@web3-ui/hooks](#web3-uihooks)
+  - [Getting started](#getting-started)
+  - [Hooks](#hooks)
+    - [useWallet](#usewallet)
+    - [useContract](#usecontract)
+    - [useTransaction](#usetransaction)
+    - [useTokenBalance](#usetokenbalance)
 
 ---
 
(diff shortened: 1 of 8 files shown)


carlomigueldy:
I am unable to test it out myself in the `example` app, it's displaying this error and I've no idea how to go around this.
```
$ yarn workspace @web3-ui/core build
$ preconstruct build
🎁 info building bundles!
🎁 success built bundles!
$ next dev
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
error - ../packages/core/src/components/ConnectWallet/ConnectWallet.tsx
Module parse failed: Unexpected token (14:4)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
|   return (
>     <>
|       <Button onClick={connected ? disconnectWallet : connectWallet}>
|         {connected
```

packages/hooks/src/hooks/useContract.ts
```diff
@@ -1,27 +1,38 @@
...
+  boolean
+];
+
+export function useContract<T extends Contract>(
+  address: string,
+  abi: ContractInterface
+): UseContractHook<T> {
   const context = React.useContext(Web3Context);
-  const [contract, setContract] = React.useState({});
+  const [contract, setContract] = React.useState<ContractInstance<T>>(null);
```
JoviDeCroock, line 25: When can this be `undefined` we're initializing it as null shouldn't we keep it stable as `T | null`?

packages/hooks/src/hooks/useContract.ts
```diff
@@ -1,27 +1,38 @@
...
-      const contractInterface = Object.values(
-        newContract.interface.functions
-      ).reduce((accumulator, funcFragment) => {
-        return {
-          ...accumulator,
-          [funcFragment.name]: newContract[funcFragment.name]
-        };
-      }, {});
-      setContract(contractInterface);
+      ) as T;
```
JoviDeCroock, line 33: Why do we have to cast this? T is extending contract so shouldn't this be inferred correctly?

JoviDeCroock (commented):
We'll need to add a changeset and document the advice for `typechain`

packages/hooks/src/hooks/useContract.ts
carlomigueldy: Ahh right, yeah it makes sense to remove `undefined` since it'll never be `undefined`

packages/hooks/src/hooks/useContract.ts
carlomigueldy: It wasn't possible and throwing the TypeScript errors though

image

image

And the `Contract` class from ethers does not have a generic type argument, so it was pretty much a work-around

carlomigueldy:
> We'll need to add a changeset and document the advice for typechain

Will write this down

packages/hooks/README.md
```diff
@@ -29,10 +29,13 @@ function MyApp({ Component, pageProps }) {
...
-- [useContract](#usecontract)
-- [useTransaction](#usetransaction)
-- [useTokenBalance](#usetokenbalance)
+- [@web3-ui/hooks](#web3-uihooks)
+  - [Getting started](#getting-started)
+  - [Hooks](#hooks)
+    - [useWallet](#usewallet)
+    - [useContract](#usecontract)
+    - [useTransaction](#usetransaction)
+    - [useTokenBalance](#usetokenbalance)
```
carlomigueldy, line 38: My VSCode extension did magic here, I reverted the changes but it was suggesting that back.

packages/hooks/README.md
```diff
@@ -76,12 +79,69 @@ import { useContract } from '@web3-ui/hooks';
...
+  'CONTRACT_ADDRESS',
+  ERC20TokenABI
+);
+
+// check that the contract has been loaded
+if (isReady) {
+  await contract.balanceOf('0x...');
+}
+```
+
```
carlomigueldy, line 144: Please proof read this and improve on parts that needs to be improved

Dhaiwat10:
@carlomigueldy are you running `yarn example` from the root?

Dhaiwat10:
Ok the example is really messed up. I'm seeing this:

```bash
yarn run v1.22.17
$ next dev
ready - started server on 0.0.0.0:3000, url: http://localhost:3000
event - compiled client and server successfully in 563 ms (1318 modules)
wait  - compiling / (client and server)...
event - compiled client and server successfully in 343 ms (1348 modules)
wait  - compiling /_error (client and server)...
event - compiled client and server successfully in 72 ms (1349 modules)
error - TypeError: stylis.middleware is not a function
```

Dhaiwat10:
Btw you can test stuff in Storybook @carlomigueldy, we have a storybook for the hooks as well. I'm gonna test this now if I can

carlomigueldy:
> @carlomigueldy are you running `yarn example` from the root?

Yes and it thrown me that error tho. I'll try again this time after deleting `node_modules` maybe got it messed up the first time since I ran `yarn dev` from `example` directory and not from root, but also realizing that I can run it from root so I ran it from root and thrown me the same error.

carlomigueldy:
> Btw you can test stuff in Storybook @carlomigueldy, we have a storybook for the hooks as well. I'm gonna test this now if I can

Oooh I see that's great, actually my first time to be using Storybook. Will also try testing it from there

carlomigueldy:
> Ok the example is really messed up. I'm seeing this:
> 
> ```shell
> yarn run v1.22.17
> $ next dev
> ready - started server on 0.0.0.0:3000, url: http://localhost:3000
> event - compiled client and server successfully in 563 ms (1318 modules)
> wait  - compiling / (client and server)...
> event - compiled client and server successfully in 343 ms (1348 modules)
> wait  - compiling /_error (client and server)...
> event - compiled client and server successfully in 72 ms (1349 modules)
> error - TypeError: stylis.middleware is not a function
> ```

Ahh I see

Dhaiwat10:
> Oooh I see that's great, actually my first time to be using Storybook. Will also try testing it from there

No problem at all. Let me know if you need any help!

packages/hooks/README.md
Dhaiwat10: Probably a good idea to revert this.

carlomigueldy: Reverted

Dhaiwat10:
Thanks @carlomigueldy ❤️  Brilliant

carlomigueldy:
> Thanks @carlomigueldy ❤️ Brilliant

Pleasure to be contributing on this project with you, legend @Dhaiwat10

PR #204, opened 2022-01-03
Draft Hooks for transactions and contract calls

### LINKED ISSUE #174: Enhance `useTransaction` hook
Currently, we are passing the function arguments while calling the `useTransaction()` hook. 
Instead of that, we should pass the function arguments while calling the method returned from this hook. 

Refer to [this](https://twitter.com/dhaiwat10/status/1475035600877092868?s=20) tweet for more reference.

Ibby-devv:
Closes #174

## Description

Modified useTransaction hook to pass the function arguments while calling the method returned from this hook. Created separate useContractFunction hook specifically for function calls.

### DIFF
--- a/packages/hooks/src/hooks/useContractFunction.ts
+++ b/packages/hooks/src/hooks/useContractFunction.ts
@@ -1,0 +1,40 @@
+import React from 'react';
+
+/**
+ * @dev Hook to get the loading status, error, and data of a function call.
+ * @param method The contract function you want to call
+ * @param args an array of arguments to pass to the function.
+ * @returns {
+ *  execute: (method, args: any[] = []) => Promise<any>,
+ *  loading: boolean,
+ *  error: null | Error,
+ * } {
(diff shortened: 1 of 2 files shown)


Ibby-devv:
I wasn't sure if this is what we were after in regards to modifying the original useTransaction hook. I noticed no provider was being passed in to the hook so I assumed it was being used in conjunction with another hook.

In this example, I'm using the Transaction Options object from ethers as a parameter.

Dhaiwat10:
@Ibby-devv we don't need a new hook, we just wanted to change the way we were passing params to the `useTransaction` hook. 😅  I'm gonna create a PR myself. Let's close this one. Sorry for getting back to you so late.

Ibby-devv:
No problems, I actually modified the one you wanted to accept the parameters how you wanted but created a separate hook purely for advanced ETH transactions which allow customisation of gas, etc.

PR #205, opened 2022-01-03
Add `useReadOnlyContract` and `useStaticProvider`

### LINKED ISSUE #202: Use static RPC providers for read-only contract functions
At the moment, the user must have connected their wallet in order for contract to be loaded and thus any contract functions to work. This shouldn't be the case. Read-only function calls don't require a signer. 

We should think of using just the `provider` for read-only functions and a `signer` for the write functions. This will probably require a drastic re-factor.

Reference: [scaffold-eth](https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/react-app/src/hooks/useStaticJsonRPC.js)

Dhaiwat10:
Closes #202 

This PR introduces the following two hooks:

- `useReadOnlyProvider` gives you a static RPC provider which you can use to read from the blockchain.
- `useReadOnlyContract`: lets you interact with a smart contract without needing a signer. (you _need_ a signer in order to use `useContract`)

Also added a new prop `rpcUrl` to the `Provider` to accept a RPC URL for the `readOnlyProvider`.

### DIFF
--- a/.changeset/metal-guests-flash.md
+++ b/.changeset/metal-guests-flash.md
@@ -2,4 +2,4 @@
 '@web3-ui/hooks': minor
 ---
 
-The `hooks` package now has two new hooks: `useReadOnlyContract` which can be used to read from contracts without a signer and `useStaticProvider`.
+The `hooks` package now has two new hooks: `useReadOnlyProvider` which gives you a read-only provider and `useReadOnlyContract` which lets you interact with a smart contract without needing a signer.
--- a/.changeset/metal-guests-flash.md
+++ b/.changeset/metal-guests-flash.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/hooks': minor
+---
+
+The `hooks` package now has two new hooks: `useReadOnlyContract` which can be used to read from contracts without a signer and `useStaticProvider`.
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -33,6 +33,8 @@ The following hooks are available:
 - [useContract](#usecontract)
 - [useTransaction](#usetransaction)
 - [useTokenBalance](#usetokenbalance)
+- [useReadOnlyContract](#usereadonlycontract)
+- [useStaticProvider](#usestaticprovider)
 
 ---
 
@@ -69,7 +71,7 @@ const {
 
 ### useContract
 
-The `useContract` hook takes the ABI and address of a contract and returns the contract instance.
+The `useContract` hook takes the ABI and address of a contract and returns the contract instance. This hook requires the user to have connected their wallet. If you don't want to force your users to connect their wallet in order to read from a contract, use [`useReadOnlyContract`](#usereadonlycontract) instead.
 
 ```tsx
 import { useContract } from '@web3-ui/hooks';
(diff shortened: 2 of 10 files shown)


packages/hooks/README.md
```diff
@@ -133,3 +135,38 @@ const {
   balanceInBigNumber
 } = useTokenBalance('TOKEN_CONTRACT_ADDRESS', 'ACCOUNT_ADDRESS');
 ```
+
+---
+
+### useReadOnlyContract
+
+The `useReadOnlyContract` hook takes in a contract address and an ABI and returns a read-only contract instance. This is especially useful when you want to read data from a function without asking the user to connect their wallet. eg. When you are only 'reading' from a contract and not interacting with it.
```
Dhaiwat10, line 143:
```suggestion
The `useReadOnlyContract` hook takes in a contract address and an ABI and returns a read-only contract instance. This is especially useful when you want to read data from a contract without asking the user to connect their wallet. eg. When you are only 'reading' from a contract and not interacting with it.
```

packages/hooks/README.md
```diff
@@ -133,3 +135,38 @@ const {
...
+
+const [contract, isReady] = useReadOnlyContract(
+  'CONTRACT_ADDRESS',
+  'CONTRACT_ABI'
+);
+```
+
+---
+
+### useStaticProvider
```
etr2460, line 162: maybe we name this `useReadOnlyProvider` too?

packages/hooks/src/hooks/useStaticProvider.ts
```diff
@@ -0,0 +1,35 @@
...
+/*
+ * @dev returns a static JSON RPC provider
+ * @param url - The url of the JSON RPC provider
+ * @credits scaffold-eth (https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/react-app/src/hooks/useStaticJsonRPC.js)
+ */
+export function useStaticProvider(url: string | undefined) {
+  const [provider, setProvider] = useState<StaticJsonRpcProvider>();
+
+  useEffect(() => {
+    async function exec() {
```
etr2460, line 20: Maybe a better name here? 
```suggestion
    async function createAndSetProvider() {
```

packages/hooks/src/hooks/useStaticProvider.ts
```diff
@@ -0,0 +1,35 @@
...
+        return;
+      }
+      try {
+        const p = await createProvider(url);
+        setProvider(p);
+      } catch (error) {
+        console.error(error);
+      }
+    }
+    exec();
```
etr2460, line 31:
```suggestion
    createAndSetProvider();
```

packages/hooks/src/hooks/useStaticProvider.ts
```diff
@@ -0,0 +1,35 @@
...
+  await p.ready;
+  return p;
+};
+
+/*
+ * @dev returns a static JSON RPC provider
+ * @param url - The url of the JSON RPC provider
+ * @credits scaffold-eth (https://github.com/scaffold-eth/scaffold-eth/blob/master/packages/react-app/src/hooks/useStaticJsonRPC.js)
+ */
+export function useStaticProvider(url: string | undefined) {
```
etr2460, line 16: why let url be undefined? It should probably be required right?
```suggestion
export function useStaticProvider(url: string) {
```

etr2460 (approved):
a couple comments, but the API seems reasonable (and good to have, connecting a wallet to unknown/untrusted sites is something we should not make the norm)

.changeset/metal-guests-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@web3-ui/hooks': minor
+---
+
+The `hooks` package now has two new hooks: `useReadOnlyContract` which can be used to read from contracts without a signer and `useStaticProvider`.
```
JoviDeCroock, line 5: Let's elaborate on the use as from what I understand the latter facilitates the former here

packages/hooks/src/hooks/useStaticProvider.ts
Dhaiwat10: So this is the issue with it: we are using this hook from `Provider.tsx` where it is optional to pass in a RPC URL which is then passed to this hook. I'm not sure how to call our hook conditionally

Dhaiwat10: Pushing a patch for now

PR #207, opened 2022-01-04
added ens lookup to address component. Requires RPC URL.

### LINKED ISSUE #200: Add ENS lookup as an option directly from Address component
By default it should not do an ENS lookup but if set to `ensLookup=True` or something similar it should be able to do the lookup. 

This would be very helpful for user-interfaces that contain list of address but don't have access to `connection.ens` for each of them. Think of a etherscan like interface or an NFT collection gallery that wants to display the ENS name of the owner if they have one.

```
export default function AddressList() {
  
const addresses =
    [
    0x54eB82E4Ec25eb173E1668dd5aB0943904d87331, 
    0x25ed58c027921E14D86380eA2646E3a1B5C55A8b, 
    0xC18360217D8F7Ab5e7c516566761Ea12Ce7F9D72
    ]

let addressList = addresses.map((address, index)=>{
    return (
        <li>
        <Address value={address} ensLookup shortened copiable />
        </li>
    )})
    return (
        <ul>
       {addressList}
       </ul>
)}
```

Other options would be add a hook that would work for this purpose or make a `<AdressList />` component

Thoughts on this?

hone1er:
Closes #200 

## 📝 Additional Information
This current solution requires an RPC and uses `provider.ensLookup()` from ethers to lookup the address. While researching how to implement this properly for looking up multiple addresses i came across this link to a `reverseRecords` [contract](https://github.com/ensdomains/reverse-records) in the [ENS domain docs](https://docs.ens.domains/dapp-developer-guide/resolving-names)

```
const namehash = require('eth-ens-namehash');
const allnames = await ReverseRecords.getNames(['0x123','0x124'])
const validNames = allnames.filter((n) => namehash.normalize(n) === n )
```

So while this is a working solution, my question is should we implement it as is, so there is an option for ens lookup directly from the component? We could also create an ENS hook and an ENS list hook. The former using ethers.provider.ensLookup for single ENS lookups, and the latter, using the `reverseRecords` contract, performing multiple address lookups.

I just want to make sure we have an efficient method for looking up multiple ENS address on a page without making a bunch of unnecessary calls.

### DIFF
--- a/.changeset/sharp-pianos-perform.md
+++ b/.changeset/sharp-pianos-perform.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/components': minor
+---
+
+Added ENS lookup directly to Address component
--- a/packages/components/src/components/Address/Address.tsx
+++ b/packages/components/src/components/Address/Address.tsx
@@ -42,7 +42,7 @@ export const Address: React.FC<AddressProps> = ({
   let feedbackTimeOut: ReturnType<typeof setTimeout>;
   let displayAddress: string = value || '';
   let [ens, setEns] = useState<string | null | undefined>(null);
-  const provider = rpcURL
+  const provider: ethers.providers.JsonRpcProvider | null = rpcURL
     ? new ethers.providers.StaticJsonRpcProvider(rpcURL)
     : null;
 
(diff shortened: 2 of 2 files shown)


Dhaiwat10:
This could be useful for looking up ENS cleanly
https://github.com/ensdomains/ui

Dhaiwat10 (approved):

PR #217, opened 2022-01-06
fix(hooks): add ERC20ABI, ERC721ABI exports

with-heart:
## Description

This PR resolves a runtime error caused by an incorrect import in `TokenGate.tsx`.

## 📝 Additional Information

Using `@web3-ui/core` caused the following error:

```
Error: Cannot find module '@web3-ui/hooks/src/constants'
```

I believe this is due to the fact that `TokenGate.tsx` was importing `ERC20ABI` using `@web3-ui/hooks/src/constants` which doesn't exist in the final build.

To resolve this, I added `ERC20ABI` as an export from `@web3-ui/hooks` and updated the imports in `TokenGate.tsx`.

### DIFF
--- a/.changeset/moody-kangaroos-melt.md
+++ b/.changeset/moody-kangaroos-melt.md
@@ -1,0 +1,6 @@
+---
+'@web3-ui/hooks': minor
+'@web3-ui/core': patch
+---
+
+Added `ERC20ABI` as an export of `@web3-ui/hooks` in order to resolve an incorrect import in the `@web3-ui/core` package.
--- a/packages/core/src/components/TokenGate/TokenGate.tsx
+++ b/packages/core/src/components/TokenGate/TokenGate.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useState } from 'react';
 import { TokenGate as PrivTokenGate } from '@web3-ui/components';
 import { useReadOnlyContract, useWallet } from '@web3-ui/hooks';
-import { ERC20ABI, ERC721ABI } from '@web3-ui/hooks/src/constants';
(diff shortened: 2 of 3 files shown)


.changeset/moody-kangaroos-melt.md
```diff
@@ -0,0 +1,6 @@
+---
+'@web3-ui/hooks': minor
```
Dhaiwat10, line 2: Shouldn't this be a patch rather than a minor bump since we are only fixing something? Would love to hear about your reasoning behind this if you think otherwise since I'm still pretty new to semantic versioning

with-heart: Nope! `core` is `minor` because we're fixing a bug. `hooks` is `minor` because the only thing we're actually doing in that package is exporting something new in the API. 

It helps to think about packages in a monorepo like they're actually in separate repos. If `@web3-ui/core` wanted to use `ERC20ABI` and it wasn't in the same repo, you'd have to add it as a new feature to `@web3-ui/hooks` first.

with-heart: Also even if it was fixing a bug in `/hooks` too, you take the highest version of the changes.

breaking change = `major`
backwards compatible functionality = `minor`
bug fix = `patch`

breaking change + bug fix = `major`
backwards compatible functionality + bug fix = `minor`

and so on!

Dhaiwat10: @with-heart thanks for writing that up. TIL!

Dhaiwat10:
@allcontributors please add @with-heart for code!

PR #218, opened 2022-01-06
chore: publish only dist folders

with-heart:
## Description

Added a `files` key to each package's `package.json` so that only the `dist` folder (and `package.json`, which is always published) is published to `npm`. 

Currently we're publishing all files in each package's directory which is unnecessary.


Dhaiwat10 (approved):
Thanks @with-heart

PR #237, opened 2022-01-11
Add `TokenBalance` component

### LINKED ISSUE #42: Token Balance Component

Dhaiwat10:
Closes #42 

## Description

This PR adds a `TokenBalance` componet to the `core` and `component` packages. `TokenBalance` fetches and renders the balance of an ERC20 token for a given account address.

### DIFF
--- a/.changeset/sixty-spiders-film.md
+++ b/.changeset/sixty-spiders-film.md
@@ -1,0 +1,6 @@
+---
+'@web3-ui/components': minor
+'@web3-ui/core': minor
+---
+
+The components and core packages now have a TokenBalance component. TokenBalance can be used to fetch and render the balance of an ERC20 token for a given account address.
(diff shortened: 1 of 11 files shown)


Dhaiwat10:
Thanks @AlexNi245 :)

PR #241, opened 2022-01-14
removed RPC, made ENS lookup optional

### LINKED ISSUE #211: [core] `Address` component that works without having to pass in provider etc

hone1er:
Closes #211 

## Description
Removed the need for an RPC URL
> Switched the need for RPC url by using window.ethereum instead
## 📝 Additional Information
This allows for the optional lookup of ENS names using a provider. It is optional because developers don't need to lookup ENS names in the component if they provide `connection.ens` and this lookup should not be used for looking up a high number of ENS names.

> For looking up many ENS names a contract called [reverseRecords](https://github.com/ensdomains/reverse-records) should be used and the names then provided to the address component.

### DIFF
--- a/packages/components/src/components/Address/Address.tsx
+++ b/packages/components/src/components/Address/Address.tsx
@@ -47,7 +47,7 @@ export const Address: React.FC<AddressProps> = ({
   let feedbackTimeOut: ReturnType<typeof setTimeout>;
   let displayAddress: string = value || '';
   const [ensName, setEnsName] = useState<string | undefined>(undefined);
-  
+
   useEffect(() => {
     if (value) {
       if (value.includes('.eth') || value === '' || value === 'Not connected')
--- a/packages/components/src/components/Address/Address.tsx
+++ b/packages/components/src/components/Address/Address.tsx
@@ -47,18 +47,16 @@ export const Address: React.FC<AddressProps> = ({
   let feedbackTimeOut: ReturnType<typeof setTimeout>;
   let displayAddress: string = value || '';
   const [ensName, setEnsName] = useState<string | undefined>(undefined);
-  const rpcProvider: ethers.providers.Web3Provider | undefined = provider
-    ? provider
-    : undefined;
+  
   useEffect(() => {
     if (value) {
       if (value.includes('.eth') || value === '' || value === 'Not connected')
         return;
     }
     async function fetchEns() {
-      if (ens && value && rpcProvider) {
+      if (ens && value && provider) {
         try {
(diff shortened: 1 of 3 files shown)


hone1er:
@Dhaiwat10 before merging...checkout this discussion on [discord](https://discord.com/channels/883478451850473483/913935677404631110/931999773651767296). We can talk about the issue here. 
But the best thing to do might be to remove the ens lookup(which is what requires the provider) and create an ENS hook instead.

packages/components/src/components/Address/Address.tsx
```diff
@@ -35,26 +35,25 @@ export const Address: React.FC<AddressProps> = ({
...
   const [error, setError] = useState<null | string>(null);
   const [copied, setCopied] = useState<boolean>(false);
   let feedbackTimeOut: ReturnType<typeof setTimeout>;
   let displayAddress: string = value || '';
-  const [ens, setEns] = useState<string | null | undefined>(null);
-  const provider: ethers.providers.JsonRpcProvider | null = rpcURL
-    ? new ethers.providers.StaticJsonRpcProvider(rpcURL)
+  const [ensName, setEnsName] = useState<string | null>(null);
+  const provider: ethers.providers.Web3Provider | null = window.ethereum
+    ? new ethers.providers.Web3Provider(window.ethereum, 'any')
```
Dhaiwat10, line 46: This will only work if there's an active wallet connection though. That's not ideal imo

hone1er: Made the changes to accept the provider as a prop and added to resolve an error i am getting while trying to use the component

packages/components/src/components/Address/Address.tsx
```diff
@@ -23,46 +27,47 @@ export interface AddressProps {
...
   let displayAddress: string = value || '';
-  const [ens, setEns] = useState<string | null | undefined>(null);
-  const provider: ethers.providers.JsonRpcProvider | null = rpcURL
-    ? new ethers.providers.StaticJsonRpcProvider(rpcURL)
-    : null;
-
+  const [ensName, setEnsName] = useState<string | undefined>(undefined);
+  const rpcProvider: ethers.providers.Web3Provider | undefined = provider
+    ? provider
+    : undefined;
```
Dhaiwat10, line 52: I feel like this line doesn't actually do anything. If someone doesn't pass in a `provider` it will be `null`. Why do we have to make it `undefined`?

hone1er: Oh yeah i suppose you're right lol. I'll go ahead and remove that

PR #249, opened 2022-01-17
refactor: Change useContract to useWriteContract

### LINKED ISSUE #219: Rename `useContract` to `useWriteContract`
Reasons why I think we should do this:

- The `useContract` hook needs a signer in order to function. It does not even load the contract if a signer isn't there. 
- In cases where we don't need a signer (eg. when we only want to read from the contract), we now have `useReadContract`.

Would love to hear everyone's thoughts if anyone has any

ngnathan:
Closes #219 

Breaking change that changes references of useContract to useWriteContract

## Description

This updates the hook `useContract` to `useWriteContract` (#219) which ensures that people that are using this hook know that it requires the user's wallet signature.
- Updated all references (README, stories, examples)

### DIFF
--- a/example/pages/index.tsx
+++ b/example/pages/index.tsx
@@ -5,24 +5,20 @@ import {
   Input,
   Button,
   Heading,
-  Text
+  Text,
 } from '@chakra-ui/react';
 import { useWallet, ConnectWallet } from '@web3-ui/core';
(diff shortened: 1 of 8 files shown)


Dhaiwat10:
@allcontributors please add @ngnathan for code!

PR #253, opened 2022-01-19
Add `TokenInput` component

### LINKED ISSUE #169: Input component with token select dropdown
### Description

Develop an input component that allows users to select tokens from the dropdown and specify number of tokens in the input.
User needs to pass an array of tokens required as props to add in the dropdown.

### Why is this needed?

This component will enable users to select a number of tokens using an easy configuration and without the requirement of passing the address of the token.

### Proposed Solution or API

Develop an input component accepting an array of required tokens to display in the dropdown. The selected token address by the user will be returned from the component.

eg: 
image

### Additional Information

_No response_

Nazeeh21:
Closes #169 

## Description
Added a `TokenInput` component.

This component will allow the user to enter the number of tokens in input and select the token from the select dropdown,

## 📝 Additional Information
Users can pass custom tokens along with their address and symbol to add as options to the select dropdown.


packages/components/src/components/TokenInput/TokenInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+export interface TokenInputProps {
+  /**
+   * an array of tokens to display
+   */
+  tokens?: Token[];
+  /**
+   * value of the input
+   */
+  value: string;
+  /**
```
Dhaiwat10, line 30: What's the difference between `value` and `selectedToken`?

packages/components/src/components/TokenInput/TokenInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+   */
+  selectedToken: string;
+  /**
+   * token change hanlder
+   */
+  onTokenChange: (token: string) => void;
+  /**
+   * want to show default tokens
+   */
+  showDefualtTokens?: boolean;
```
Dhaiwat10, line 49:
```suggestion
  showDefaultTokens?: boolean;
```

Dhaiwat10: Please also update any references to this in this file or other files

Dhaiwat10:
Can you add a story that demonstrates a custom `width`?

packages/components/src/components/TokenInput/TokenInput.tsx
Nazeeh21: - `selectedToken` will contain the address of the selectedToken
- `value` will be the value of the input element

packages/components/src/components/TokenInput/TokenInput.tsx
Nazeeh21: Yeah, sure

Nazeeh21:
> Can you add a story that demonstrates a custom `width`?

I've added a story that adds allows adding a border to the whole `TokenInput` element and with custom width.

packages/components/src/components/TokenInput/TokenInput.tsx
Dhaiwat10: @Nazeeh21 I think `amount` is a better name for the prop than `value`

Dhaiwat10: You should also rename `onValueChange` to `onAmountChange`

Dhaiwat10:
@Nazeeh21 we also need to add some basic docs around this component to the Readme. Can you take care of that too?

Nazeeh21:
> @Nazeeh21 we also need to add some basic docs around this component to the Readme. Can you take care of that too?

Yeah sure!

packages/components/src/components/TokenInput/TokenInput.tsx
Nazeeh21: Okay, I'll do that

.changeset/stupid-rings-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@web3-ui/components': minor
+---
+
+Add TokenInput component
```
with-heart, line 5: Since this is added to the changelog, it'd be great to add more detail so our users understand _what_ this component is, _why_ it's useful, and _how_ they can use it.

````suggestion
Added the `TokenInput` component which combines an `input` for the user to provide an amount of tokens and a `select` for the user to select the type of token. 

```tsx
const Example = () => {
  const [value, setValue] = React.useState('');
  const [token, setToken] = React.useState('');
  
  return (
    <TokenInput
    	tokens={['BTC', 'ETH', 'GTC']}
      value={value}
      onValueChange={event => setValue(event.target.value)}
      selectedToken={token}
      onTokenChange={event => setToken(event.target.value)}
    />
  )
}
```
````

**Comment by with-heart** [comment]
There are a few issues here that make me think we might want to consider making this a [compound component](https://www.smashingmagazine.com/2021/08/compound-components-react/) with state shared via `Context` rather than a single component:

- The presence of `inputProps` and `selectProps`. Needing to pass objects of props that are forwarded to other children is usually a sign that a single component is not flexible enough.
- Needing to specify the `select`'s options using an array prop rather than using `jsx`. This means there's not an easy way for a user to display options where the text is different from the token's ``symbol` and `value` is different from the token's `address`.

Let's look at an example. The current way:

```tsx
const tokens = [
  { symbol: 'ETH', address: '0xethaddress' },
  { symbol: 'GTC', address: '0xgtcaddress' },
]

const Example = () => {
  const [value, setValue] = React.useState('');
  const [token, setToken] = React.useState('');

  return (
    <TokenInput
      width="fit-content"
      border="2 px solid black"
      borderWidth="2px"
      p={2}
      tokens={tokens}
      selectedToken={token}
      onTokenChange={setToken}
      value={value}
      onValueChange={setValue}
      inputProps={{
        width: '20rem',
      }}
      selectProps={{
        borderLeft: '2px solid black',
        fontSize: '1.2rem',
        fontWeight: 'bold',
        w: '7rem',
        outline: 'none',
        rounded: 'none',
        _focus: {
          borderLeft: '2px solid black',
        },
      }}
    />
  );
};
```

With compound components, we could instead do something like this:

```tsx
const Example = () => {
  const [value, setValue] = React.useState('');
  const [token, setToken] = React.useState('');

  return (
    <TokenInput
      width="fit-content"
      border="2 px solid black"
      borderWidth="2px"
      p={2}
      selectedToken={token}
      onTokenChange={setToken}
      value={value}
      onValueChange={setValue}
    >
      <TokenInput.Input width="20rem" />
      <TokenInput.Select
        borderLeft="2px solid black"
        fontSize="1.2rem"
        fontWeight="bold"
        w="7rem"
        outline="none"
        rounded="none"
        _focus={{ borderLeft: '2px solid black' }}
      >
        <option value="ETH">ETH</option>
        <option value="GTC">GTC</option>
      </TokenInput.Select>
    </TokenInput>
  );
};
```

packages/components/src/components/TokenInput/TokenInput.tsx
```diff
@@ -0,0 +1,110 @@
...
```
with-heart, line 50: I think we should probably just let users provide their own tokens. There's a lot of bikeshedding that could happen here:

- Why are these tokens included as defaults but not others?
- What if I want all of the default tokens except one of them?

At the very least, these should be documented in the `jsdoc` string so users know what they're going to get by enabling this.

packages/components/src/components/TokenInput/TokenInput.tsx
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+export const TokenInput: React.FC<TokenInputProps & BoxProps> = ({
+  tokens = [],
+  value,
+  onValueChange,
+  label,
+  selectedToken,
+  onTokenChange,
+  showDefualtTokens = true,
```
with-heart, line 72: If we keep this prop, we should probably default it to `false`. imo it'd be unexpected and confusing to render a `select` and have `options` appear that you didn't specify.

Nazeeh21:
@with-heart I really like your idea of compound components. I also felt the passing props in such a manner is not the optimized way but couldn't find the way around. But thinking about the compound component, I feel that's the best UX we can provide with the same functionality.

.changeset/stupid-rings-pump.md
Nazeeh21: Agreed, I'll add more info in the upcoming commits

Dhaiwat10:
@Nazeeh21 will you be working on making those changes? Let me know if you don't have the bandwidth, no pressure

Nazeeh21:
@Dhaiwat10 thanks for asking. Yes, i feel that I've much things rn on my plate and making these changes will require me upto 3 days. So i think it'd be really nice if someone can come forward to make these changes.

with-heart:
I'd be happy to work on it. Even happier if we could pair on it @Nazeeh21!

Nazeeh21:
@with-heart I've tried creating Compound Components. Please have a look and LMK if any of the changes are required.

packages/components/src/components/TokenInput/TokenInput.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    </TokenInput>
+  );
+};
+
+describe('TokenInput', () => {
+  it('renders correctly', () => {
+    const { container } = render(<Component />);
+    expect(container).toBeInTheDocument();
+  });
+});
```
Dhaiwat10, line 41: Can you add more tests? Something that verifies that the tokens are actually getting selected correctly, etc?

.changeset/stupid-rings-pump.md
```diff
@@ -0,0 +1,22 @@
...
+Added the `TokenInput` component which combines an `input` for the user to provide an amount of tokens and a `select` for the user to select the type of token.
+
+```tsx
+const Example = () => {
+  const [value, setValue] = React.useState('');
+  const [token, setToken] = React.useState('');
+
+  return (
+    <TokenInput
+      tokens={['BTC', 'ETH', 'GTC']}
```
Dhaiwat10, line 14: This does not look like the right API. Aren't you supposed to pass in an array of objects?

packages/components/src/components/TokenInput/TokenInput.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+        }}
+        token={token}
+        setToken={setToken}
+        tokensList={defaultTokens}
+      />
+    </TokenInput>
+  );
+};
+
+const Component3 = () => {
```
Dhaiwat10, line 46: You should give these better names to improve readability

packages/components/src/components/TokenInput/tokens.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+  {
+    symbol: 'GTC',
+    address: '0xde30da39c46104798bb5aa3fe8b9e0e1f348163f',
+  },
+  {
+    symbol: 'USDC',
+    address: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48',
+  },
+];
```
Dhaiwat10, line 24: Let's not include any default tokens at all. Make `tokens` a required prop please.

packages/components/src/components/TokenInput/tokens.ts
```diff
@@ -0,0 +1,24 @@
...
+    symbol: 'USDT',
+    address: '0x55d398326f99059fF775485246999027B3197955',
+  },
+  {
+    symbol: 'DAI',
+    address: '0x6B175474E89094C44Da98b954EedeAC495271d0F',
+  },
+  {
+    symbol: 'ETH',
+    address: '0xc3761eb917cd790b30dad99f6cc5b4ff93c4f9ea',
```
Dhaiwat10, line 14: ETH is not an ERC20 token 😅

Dhaiwat10 (changes requested):

packages/components/src/components/TokenInput/TokenInput.test.tsx
Nazeeh21: Yeah definitely, I'll add

.changeset/stupid-rings-pump.md
Nazeeh21: Yes exactly, I'll have a look to it.

packages/components/src/components/TokenInput/tokens.ts
Nazeeh21: Yes, I've updated the component to accept the `tokens` as a required prop. I've kept this to use in the stories.
Also, I've updated the `tokens` prop to `tokenList` prop.

packages/components/src/components/TokenInput/tokens.ts
Nazeeh21: Yeah, I'll remove it 😅

Nazeeh21:
@Dhaiwat10 let me know if any of the changes are required on this, :)

Dhaiwat10:
Hey @with-heart can you please quickly review this so that we can merge this before I start the re-work? Thanks!

Nazeeh21:
@with-heart thanks for your comment. I was in Dubai the previous week and am just finishing my backlog about some stuff. I've skimmed through your comment but have not read it in detail. I'll come back to it in 2-3 days. Sorry for the delay.

Nazeeh21:
> I'm wondering if this is even that useful of an abstraction. The only thing it's really doing is wrapping up a few Chakra components, setting w="fit-content" on two of them, and mapping tokensList to option elements. In some ways this feels like more of a recipe than a needed & useful component. That doesn't necessarily mean we shouldn't use it though.

@with-heart  After creating the compound component, I was also feeling the same. As the development on this going to make a fresh start and we are no more using Chakra, I think this is the right time to give a second thought to this component whether we really want it or not. 
Would like to hear your thoughts as well on this @Dhaiwat10.

I've noted all the above suggestions. If we'll decide to keep it, I'll def make the suggested changes. Thanks

Issue #175, opened 2021-12-26
Add tests for our hooks


Dhaiwat10:
None of our hooks have tests at the moment. Not good. Let's add tests so that we know when some shit breaks.

- [x] #176
- [ ] #177
- [x] #178
- [ ] #179

PR #258, opened 2022-01-20
Feat/175 use token balance tests

AlexNi245:
Closes # #175 

## Description
This pull request adds tests for the useTokenBalance hook. That is possible because we're now having a way to mock ethers and hence don't need to interact with the blockchain at all. Thanks to @with-heart for figuring this out with me together.

I think this PR will really be useful for other developers which are willing to write tests for their hooks. So I think it would make sense to share this knowledge in the README.md or maybe provide some skeleton test file. I would love to hear your ideas on this.

## Aditional information

```

jest.mock('ethers', () => ({
  ...jest.requireActual('ethers'),
  ethers: {
    ...jest.requireActual('ethers').ethers,
    Contract: class ContractStub {
      balanceOf(addr) {
        //Throws an exception if address is not valid
        //And causes the hook to fail
        ethers.utils.getAddress(addr);
        return BigNumber.from(ethers.utils.parseEther('1.0'));
      }
      decimals() {
        return ethers.BigNumber.from('18');
      }
    },
  },
}));
```

This snippet is used to mock the ethers.js node_module. In this example, we're mocking the Contract class and their methods decimals and balanceOf(). You can use this to mock whatever class from ethers.js you want to mock.


packages/hooks/src/hooks/useTokenBalance.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+      }
+    },
+  },
+}));
+
+describe('useTokenBalanceTest', () => {
+  test('happy path - returns accountsAddresses balance if both accountAddress and tokenAddress are defined  ', async () => {
+    const wrapper = ({ children }) => (
+      <>
+        {/* @ts-ignore */}
```
AlexNi245, line 29: I've added ts-ignore on purpose, so we can pass a mocked provider object. We need this to pass undefined checks in the actual hook. However, the provider is never used because we are mocking ethers.js so how to tell the compiler that it's fine to not pass an actual implementation of provider?

with-heart:
I did a bit of cleanup for this:

- merged `waitFor` calls for each test into a single call that checks the entire `result.current` object. makes things a little cleaner
- created a single `Web3ContextTestProvider` to use as `wrapper` for all tests rather than a separate `wrapper` for each test
- added comments explaining the mocking situation, just for future viewers to understand
- replaced `@ts-ignore` with `@ts-expect-error` (required by our `eslint` rules) and removed one unused `@ts-ignore`
- removed the `changesets` file. these messages are added to our public `CHANGELOG` for our end users and testing is irrelevant to them. we only need to add them for bug fixes, new features, etc.

packages/hooks/src/hooks/useTokenBalance.ts
```diff
@@ -33,25 +33,29 @@ export function useTokenBalance({ tokenAddress, accountAddress }: Props) {
...
-    if (tokenAddress && accountAddress && provider) {
-      setError(false);
-      setLoading(true);
-      try {
-        getBalance();
-      } catch (error) {
-        console.error(error);
-        setError(true);
-      }
+    if (!(tokenAddress && accountAddress && provider)) {
```
Dhaiwat10, line 50: @AlexNi245 can you describe why you made these changes? Once this is out of the way, we can merge.

Dhaiwat10:
Thanks @AlexNi245 :) sorry for the late review, but left one question.

packages/hooks/src/hooks/useTokenBalance.ts
AlexNi245: Yes, I made the changes to ensure that isLoading is set to false and isError is set to true if one of the 3 conditions in line 50 is not satisfied. In my opinion, it is a bit more readable to return if the conditions are not true rather than handle the error in the else block.

PR #274, opened 2022-01-29
Relax network attribute for Provider component

juliankrispel:
## Description

This pr simply relaxes the type constraint for the network prop on the `Provider` component.

Although the type currently says that it is a required prop it is not in actually and simply a different but valid configuration of this component. 

## 📝 Additional Information

Rendering a provider like this `<Provider></Provider>` results in having the network inferred from the wallet, which is afaik a totally valid use-case.

### DIFF
--- a/packages/components/README.md
+++ b/packages/components/README.md
@@ -18,7 +18,7 @@ import { NETWORK } from '@web3-ui/core';
 // Passing in a theme is optional
 function MyApp({ Component, pageProps }) {
   return (
-    <Provider theme={yourChakraTheme} network={NETWORKS.mainnet}>
+    <Provider theme={yourChakraTheme}>
       <Component {...pageProps} />
     </Provider>
   );
--- a/packages/components/README.md
+++ b/packages/components/README.md
@@ -13,7 +13,6 @@ At the root of your React app, wrap your app in the `<Provider>` component:
 ```tsx
 // _app.tsx (or the root of your app)
 import { Provider } from '@web3-ui/components';
-import { NETWORK } from '@web3-ui/core';
 
 // Passing in a theme is optional
 function MyApp({ Component, pageProps }) {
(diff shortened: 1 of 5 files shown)


packages/components/README.md
```diff
@@ -13,11 +13,12 @@ At the root of your React app, wrap your app in the `<Provider>` component:
 ```tsx
 // _app.tsx (or the root of your app)
 import { Provider } from '@web3-ui/components';
+import { NETWORK } from '@web3-ui/core';
 
 // Passing in a theme is optional
 function MyApp({ Component, pageProps }) {
   return (
-    <Provider theme={yourChakraTheme}>
+    <Provider theme={yourChakraTheme} network={NETWORKS.mainnet}>
```
Dhaiwat10, line 21: You can't pass in a `network` prop to a `Provider` imported from the `components` package. We should revert changes made to this file 🙂

Dhaiwat10:
> Rendering a provider like this <Provider></Provider> results in having the network inferred from the wallet, which is afaik a totally valid use-case.

Interesting. I agree and am totally fine with this but just want to make sure that nothing is breaking. Did you test this out?

juliankrispel:
> > Rendering a provider like this  results in having the network inferred from the wallet, which is afaik a totally valid use-case.
> 
> Interesting. I agree and am totally fine with this but just want to make sure that nothing is breaking. Did you test this out?

@Dhaiwat10 yes we're using it like this hence the pr.

packages/components/README.md
juliankrispel: I did but - it seems like that statement isn't true 🤔 

the Provider component has a network prop in components as well as in hooks

image

Dhaiwat10 (approved):
Cool! LGTM! Thanks @juliankrispel

Issue #245, opened 2022-01-16
Add CRA Example


codingwithmanny:
**Context:**

After trying to setup `web3-ui` locally with [CRA](https://create-react-app.dev).

**Proposal:**

Organize the examples folder to separate the different react platforms you can use to serve react with.

**_NOTE:_** This is just CRA as an example and then adding more and more in different issues/pull requests.

_Example:_ Proposed folder structure.

```diff
- example
+ examples
+  - nextjs
+  - vitejs
+  - cra
+  - gatsby
```

**Goal:**

Give more examples of how to use `web3-ui` with different libraries and frameworks.

**Additional Details:**

Related to #242

codingwithmanny:
@e-roy Has mentioned he would like to take this on.

e-roy:
I'll replicate the ui from the [PR](https://github.com/Developer-DAO/web3-ui/pull/244) when it gets merged if that works

codingwithmanny:
@e-roy you still up for the CRA implementation?

e-roy:
@codingwithmanny so I created a  [fork](https://github.com/e-roy/web3-ui/tree/cra-example)

I have a few questions on this before creating a PR

1. Do the examples need to be in the workspaces?  
I had trouble getting packages to load properly using workspaces.  I noticed both of the other examples had yarn.lock files.

2. I replicated the vitejs example for this.  The only difference is in the vite example, useWriteContract was used from the core package and I used useContract from the hooks package.  It looks like the change to useWrieContract from [refactor](https://github.com/Developer-DAO/web3-ui/commit/1bb2aa67f533a8699c4dcd14381a71fb7e495c7f#diff-37008fa4a176034754370e62d3a5c1a93cc0376e004b1a7434ecb7b9bcaea02a) isn't in the newest npm, so I used the older useContract.  

What are your thoughts on this?

e-roy:
closing because of massive structure change

PR #275, opened 2022-01-31
CRA example

e-roy:
Closes # [245](https://github.com/Developer-DAO/web3-ui/issues/245)

## Description

This is a CRA example using the web3-ui library.  This example is based off the vitejs example.

## 📝 Additional Information

Removed the examples folder from the workspaces.  When using the examples folder within workspaces, the web3-ui library could not be found.  I removed the line in the root package.json file for examples in workspaces so this example would reflect a user copying this folder as a starting point.


Dhaiwat10:
Hey thanks @e-roy! 

- I think we should rename the folder to `create-react-app` instead from `reactjs`
- When I run `yarn` from the root, it does not also install the dependencies for this example. You can fix this by adding this folder to the yarn workspace that we have.

The example works fine other than that. :)

e-roy:
@Dhaiwat10 sorry for the delayed reply, I wanted to double check this.

I changed the folder name to `create-react-app`

This PR actually removes the examples folder from the workspaces.  When I had the examples folder included and I ran `yarn` from the root, I would get a list of errors, attached picture:

web3-ui-react-errros

I'm not sure what I can do resolve this so that packages are installed properly by running `yarn` from the root folder instead of the example's folder.

examples/create-react-app/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "react-app",
+  "version": "0.1.0",
```
with-heart, line 3:
```suggestion
  "version": "0.1.0",
  "private": true,
```

We need this to be `private` to ensure that it can't be published

with-heart (changes requested):

with-heart:
I think the `useContract` errors are because that hook doesn't exist anymore. It was renamed to `useWriteContract` in #249 (though that didn't get a `changeset` file so it's not listed in a `CHANGELOG` 😭)

e-roy:
@with-heart thanks for catching that!

I'm using"@web3-ui/core": "^0.5.1" in my example.  When I tried the newer 0.5.2 version I got this error.  

web3-ui-error


I wasn't sure if it was something I was doing wrong, but I'm seeing the same in a different project I'm testing with.  I'm thinking I should open a bug issue for this?

Dhaiwat10:
> I wasn't sure if it was something I was doing wrong, but I'm seeing the same in a different project I'm testing with. I'm thinking I should open a bug issue for this?

Please file an issue. Looks like something clearly broke lol. And that was me who forgot to include the renaming with a changeset sorry guys

with-heart:
> > I wasn't sure if it was something I was doing wrong, but I'm seeing the same in a different project I'm testing with. I'm thinking I should open a bug issue for this?
> 
> Please file an issue. Looks like something clearly broke lol. And that was me who forgot to include the renaming with a changeset sorry guys

Issue: #288 
Resolution: #289

with-heart (commented):
We'll have to wait till a new version of `@web3-ui/hooks` is published to switch back to `useWriteContract` and get this merged

e-roy:
@with-heart this will solve a concern of having a React example that is using the newer `useWriteContract` hook, but doesn't address the initial concern @Dhaiwat10 had with 

> When I run yarn from the root, it does not also install the dependencies for this example. You can fix this by adding this folder to the yarn workspace that we have. 

Which I addressed with 

> @Dhaiwat10 sorry for the delayed reply, I wanted to double check this.
> 
> I changed the folder name to `create-react-app`
> 
> This PR actually removes the examples folder from the workspaces. When I had the examples folder included and I ran `yarn` from the root, I would get a list of errors, attached picture:
> 
> web3-ui-react-errros
> 
> I'm not sure what I can do resolve this so that packages are installed properly by running `yarn` from the root folder instead of the example's folder.

The screen shot of errors only shows a few of all the errors.  If you would like all the errors please let me know and I'll get every one.  But the most of the errors other than `useContract`, are around `Module parse failed: Unexpected token`

Removing the examples folder from workspaces, isolates the examples from the rest of the repo.  If it's intended for a user to go to the example's folder and copy the contents of the framework of choice to get up and running quickly, isolating the examples by NOT including in workspaces, would be the best way.  This way each example framework has their own package.json file with it's own working version of web3-ui, and a user copying would not experience future breaking changes to library updates.

If this is the desired method, than this PR is good to merge because it uses v 0.5.1 that uses `useContract`.  When the newer working version is published, than a separate issue could be raised to update this example to the newer web3-ui package version.  

If this is NOT the desired method, and you want to have the example's folder included in the root package.json workspaces, then the multiple errors around `Module parse failed: Unexpected token` still needs to be solved.  I was not able to figure it out.

Dhaiwat10:
> We'll have to wait till a new version of @web3-ui/hooks is published to switch back to useWriteContract and get this merged

Just published new versions for all three packages! Thanks for the help :)

Dhaiwat10:
> If this is NOT the desired method, and you want to have the example's folder included in the root package.json workspaces, then the multiple errors around Module parse failed: Unexpected token still needs to be solved. I was not able to figure it out.

I think we should continue the convention we have had for our other examples and try to find a solution for this module parse error. I unfortunately don't have a lot of time right now but I will try my best to do some digging when I can. I appreciate you taking the time to write this out and trying to find a solution. Thanks man

Dhaiwat10:
I'm totally open to hear any other suggestions though, above is just my opinion!

e-roy:
> I'm totally open to hear any other suggestions though, above is just my opinion!

no worries, you are the maintainer :)

> I think we should continue the convention we have had for our other examples and try to find a solution for this module parse error. I unfortunately don't have a lot of time right now but I will try my best to do some digging when I can. I appreciate you taking the time to write this out and trying to find a solution. Thanks man

I believe the nextjs example is having the same troubles.  Maybe @codingwithmanny has some insight how he was able to get the example of vitejs working in this convention?

codingwithmanny:
@e-roy will take a look tomorrow

Dhaiwat10:
haha @codingwithmanny can you have a quick look at this one so we can merge before I start the re-work? :)

codingwithmanny:
@Dhaiwat10 @e-roy Looking at this again, I think there are some changes that web3-ui would need to solve because I don't rewire is the route to go to say that a library requires a user to reconfigure their CRA vs it being out of the box to support a library makes a difficult sell for a dev to adopt.

Dhaiwat10:
@codingwithmanny that's fair. Let's come back to this after I re-work the project

Dhaiwat10:
@allcontributors please add @e-roy for code!

e-roy:
> @Dhaiwat10 @e-roy Looking at this again, I think there are some changes that web3-ui would need to solve because I don't think rewire is the route to go. To say that a library requires a user to reconfigure their CRA vs it being out of the box to support a library makes a difficult sell for a dev to adopt.

@Dhaiwat10 and @codingwithmanny some thoughts 

I'm not a big fan of using rewire either, especially for when trying to create an out of the box solution for a newer dev.  I don't think there can be changes to web3-ui that will avoid using something like rewire.   I believe issue centers around changes in [Webpack 5](https://webpack.js.org/configuration/resolve/#resolvefallback).  I've had to do something very similar to this when using the wagmi library and also using moralis sdk, in react, in other projects.  

[with moralis](https://github.com/e-roy/nft-dashboard-template/tree/main) - p.s. from a hackathon, I was asked to iterate off another repo, it became a bit sloppy.
[with web3-ui](https://github.com/e-roy/react-app-web3-ui-start) - just more context to the conversation.
[wagmi](https://github.com/tmm/wagmi/tree/main/examples) - just showing wagmi doesn't have an example either.  Sorry I don't have a public repo with react and wagmi to show.

From what I'm seeing around, I think this is happening across the board with web3 libraries and react with webpack 5.  Not sure if this will change with react 18

PR #276, opened 2022-01-31
feature: Receive any argumentss on execute method from useTransaction

### LINKED ISSUE #261: Switch array `args` to `arguments` on `execute` from `useTransaction`
It's easier to call this function as we were calling the original function.

Current implementation:

```javascript
await execute([1,
	{
		value: ethers.utils.parseEther("0.05"),
	}
])
```

Suggested implementation:

```javascript
await execute(1,
	{
		value: ethers.utils.parseEther("0.05"),
	}
)
```

berteotti:
Closes #261 

## Description

> Update `execute` method from `useTransaction` to receive any arguments.

## 📝 Additional Information

> We should call `execute` as we were calling the original method instead of an array as first argument.

### DIFF
--- a/.changeset/strong-suns-sort.md
+++ b/.changeset/strong-suns-sort.md
@@ -1,0 +1,12 @@
+---
+"@web3-ui/hooks": minor
+---
+
+You can now pass in args to the exec fn in useTransaction just like you would with ethers. You don't need to pass it all inside an array anymore.
+
+```tsx
+execute(['Hello', { value: '10' }]);
+
+// is now
+execute('Hello', { value: '10' });
+```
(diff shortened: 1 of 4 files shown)


Dhaiwat10:
@berteotti thanks! One last thing: can you add a minor changeset to the hooks package (run `yarn changeset`) with the following message: 

You can now pass in args to the exec fn in useTransaction just like you would with ethers. You don't need to pass it all inside an array anymore.

```tsx
execute(['Hello', { value: '10' }]);

// is now
execute('Hello', { value: '10' });
```

Dhaiwat10:
I tried pushing a changeset but it isn't showing up for some reason

berteotti:
> I tried pushing a changeset but it isn't showing up for some reason

@Dhaiwat10 can you confirm that's the desired changeset please?

Dhaiwat10:
@berteotti yep that's perfect, thanks

Dhaiwat10:
@allcontributors please add @berteotti for code!

PR #283, opened 2022-02-01
Add alert if network is not added in metamask

### LINKED ISSUE #259: switchToCorrectNetwork fails if the requiredNetwork is moonriver
When metamask is set to mainnet, and I want to switch to moonriver (chainId 1285), I'm getting error code 4902. This is most likely because a different [chainId/rpcUrl](https://github.com/Developer-DAO/web3-ui/blob/43d7031185f683e9e64329ba020bd721502dc513/packages/hooks/src/hooks/useWallet.ts#L58) is being added to handle the error: https://github.com/Developer-DAO/web3-ui/blob/43d7031185f683e9e64329ba020bd721502dc513/packages/hooks/src/hooks/useWallet.ts#L55

NOTE: moonriver network is already added to my metamask browser extension

Nazeeh21:
Closes #259 

## Description

Display alert if the user doesn't have the required network installed.

In the future, we can integrate adding the required network to the user's metamask.

### DIFF
--- a/packages/hooks/src/hooks/useWallet.ts
+++ b/packages/hooks/src/hooks/useWallet.ts
@@ -23,12 +23,12 @@ export function useWallet() {
     provider,
     correctNetwork,
     network,
-    readOnlyProvider
+    readOnlyProvider,
   } = context;
 
   React.useEffect(() => {
     if (userAddress && provider && chainId === NETWORKS.mainnet) {
-      provider.lookupAddress(userAddress).then(address => {
+      provider.lookupAddress(userAddress).then((address) => {
         setEns(address as string);
       });
     }
@@ -42,7 +42,7 @@ export function useWallet() {
         // check if the chain to connect to is installed
         await window.ethereum.request({
           method: 'wallet_switchEthereumChain',
-          params: [{ chainId: `0x${network}` }] // chainId must be in hexadecimal numbers
+          params: [{ chainId: `0x${network}` }], // chainId must be in hexadecimal numbers
         });
       } catch (error) {
         console.error(error);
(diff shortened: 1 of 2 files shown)


packages/hooks/src/hooks/useWallet.ts
```diff
@@ -50,19 +50,7 @@ export function useWallet() {
...
-                {
-                  chainId: '0x61',
-                  rpcUrl: 'https://data-seed-prebsc-1-s1.binance.org:8545/',
-                },
-              ],
-            });
-          } catch (addError) {
-            console.error(addError);
-          }
+          alert('Please add the network to your MetaMask');
```
juliankrispel, line 53: `alert` is jarring UX, if we handle this shouldn't we handle it properly? How about setting a state variable and exposing it?

juliankrispel (commented):
Great work. Only thing - `alert` makes for pretty bad UX and is therefore not really used. Could we just remove that and log the error until there's a better solution?

packages/hooks/src/hooks/useWallet.ts
Nazeeh21: Yeah alert kinda annoy users, so I think returning the error to the user or logging it to the console would work. WDYT @Dhaiwat10 ?

Dhaiwat10: The problem with only logging it is that non-technical users won't even know that there's an error. Alert is ugly but atleast it lets them know about what went wrong.

Nazeeh21:
Okay, so if no additional changes are required and everything seems alright then let's merge this. :)

PR #292, opened 2022-02-05
Fix/issue 230

### LINKED ISSUE #230: Replace uses of `null` with `undefined`
> I call it my billion-dollar mistake. It was the invention of the null reference in 1965.

— [Sir Tony Hoare](https://en.wikipedia.org/wiki/Null_pointer)

JavaScript unfortunately has two billion-dollar mistakes: `null` and `undefined`.

While we can't remove them from the language, we can reduce our usage down to a single type. I'd like to propose that we replace all uses of `null` with `undefined`.

The reason I suggest `undefined` is due to its natural existence through JavaScript and TypeScript. For example, check out this TypeScript function with an optional argument:

```ts
function useOptionalArg(optionalArg?: string) {
  console.log(optionalArg)
}
```

The use of the `?` operator makes `optionalArg` have a type of `string | undefined`.

Additionally, using only `undefined` allows us to avoid awkward type situations like this one:

https://github.com/Developer-DAO/web3-ui/blob/f481a3670ec92a7653772b8ee5be3123a7641eb6/packages/hooks/src/hooks/useTokenBalance.ts#L13-L22

The type of `provider` on line 22 is `ethers.providers.Web3Provider | null | undefined`. This is due to the way `provider` is defined on `Web3Context`:

https://github.com/Developer-DAO/web3-ui/blob/f481a3670ec92a7653772b8ee5be3123a7641eb6/packages/hooks/src/Provider.tsx#L16

There's no reason for it to also be typed as `null` when it's already potentially `undefined` due to it being optional.

dfimbres:
Closes #230 

## Description
Replaced values of null to undefined.

### DIFF
--- a/example/pages/index.tsx
+++ b/example/pages/index.tsx
@@ -15,7 +15,7 @@ import GreeterABI from '../abis/Greeter.json';
 
 export default function Home() {
   const [address, setAddress] = useState('');
-  const [nftGallery, setNftGallery] = useState(null);
+  const [nftGallery, setNftGallery] = useState(undefined);
   const { correctNetwork, switchToCorrectNetwork, connected, provider } =
     useWallet();
   const [greeterContract, isReady] = useWriteContract<Greeter>(
--- a/packages/components/src/components/Address/Address.tsx
+++ b/packages/components/src/components/Address/Address.tsx
@@ -37,14 +37,14 @@ export const Address: React.FC<AddressProps> = ({
   shortened = false,
   rpcURL,
 }) => {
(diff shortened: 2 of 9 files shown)


Dhaiwat10:
@allcontributors please add @dfimbres for code

Dhaiwat10:
@dfimbres since we are now plannig to rework the entire project (context: https://twitter.com/dhaiwat10/status/1499264565816999940?s=20), it is best if we close this PR. I appreciate you creating this PR though, I really do. Please feel free to create another PR once I create more issues for the rework!

PR #293, opened 2022-02-06
Move `ethers` to peer dependencies

### LINKED ISSUE #220: Move `ethers` to peer dependencies
After some discussion, the team has some consensus over moving `ethers` to peer dependencies instead of having it as a dependency. [Read the discussion here](https://discord.com/channels/883478451850473483/913935677404631110/928279789184364574)

When we work on this issue, we should also remember to add some documentation around this. We should especially tell our users to install `ethers` alongside our packages in our `yarn add ...` snippets in the Readmes.

Dhaiwat10:
Closes #220 

### ! **BREAKING CHANGE** !
This PR moves `ethers` to the peer dependencies for the `components` and `hooks` package.

This means that anyone who uses our packages will now also have to manually install `ethers` alongside our packages.

Why? Mainly to avoid dependency conflicts in cases when people already have `ethers` installed.

### DIFF
--- a/.changeset/khaki-cobras-laugh.md
+++ b/.changeset/khaki-cobras-laugh.md
@@ -1,0 +1,9 @@
+---
+'@web3-ui/components': minor
+'@web3-ui/hooks': minor
+---
+
+!! BREAKING CHANGE !!
+`ethers` has been moved to peer dependencies for both the `components` and `hooks` packages. (It was already a peer dependency for `core`)
+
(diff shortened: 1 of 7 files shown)


with-heart (approved):
Looks good to me!

with-heart:
Actually having seen @JoviDeCroock's messages in `#open-source-team`, I think we should consider refining the `peerDependencies` a bit before merging!

Dhaiwat10:
@with-heart I agree, will get back to this as soon as I can

PR #297, opened 2022-02-10
fix: save chainId on chainChanged event without formatting

### LINKED ISSUE #273: Chain Id Formatted Incorrectly When Changing Networks
When changing networks/blockchains the Web3 context incorrectly updates the Chain ID from hexadecimal. For example, the localhost chain Id is stored as 1337 in decimal, but 0x539 in hex. The current formatting for 'chainChanged' just rips the 0x off the front instead of a conversion to decimal. 

Example: If you are on mainnet and switch to localhost, the chainId in the Web3 context will update to 539 instead of 1337 and there will be a mismatch between the network and chainId and correctNetwork will be false. 

In the 'accountsChanged' listener right below it, the chainId is set from the ethers provider and seems to have no issues. 

See screenshots below. I'm suggesting using the method marked by red 2 in the listener marked by the red 1. Apologies for any confusion in the explanation. 

Once I can get my environment set up properly I'd be happy to do this. Thanks :). 

chainIdProps
web3ui issue

shamoilarsi:
Closes #273 

## Description

removed chainId formatting from `chainChanged` event. 
before -> chainId `0x89` would be saved as `89`.
after -> chainId `0x89` will be saved as `137`

### DIFF
--- a/.changeset/curvy-phones-agree.md
+++ b/.changeset/curvy-phones-agree.md
@@ -1,0 +1,6 @@
+---
+"@web3-ui/hooks": patch
+---
+
+closes #273
+fix: save chainId on chainChanged event without formatting
--- a/packages/hooks/src/Provider.tsx
+++ b/packages/hooks/src/Provider.tsx
@@ -119,7 +119,7 @@ export const Provider: React.FC<ProviderProps> = ({
       setConnected(true);
 
       connection.on('chainChanged', async (newChainId: string) => {
-        const formattedChainId = +newChainId.split('0x')[1];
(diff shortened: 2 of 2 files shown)


shamoilarsi:
@Dhaiwat10 I've tested this by logging the chainId and it seems to be saved correctly. 
let me know if there is some other way I should test this

Dhaiwat10:
Thanks @shamoilarsi! ❤️

Dhaiwat10:
@allcontributors please add @shamoilarsi for code

PR #311, opened 2022-02-24
Feat/network switch component

### LINKED ISSUE #247: [core] Network switch component
### Description

We should have a network switch similar to the one seen at [Uniswap](https://app.uniswap.org/#/swap) that lets people easily switch between networks. 

This network switch component should automatically send a network change request to the users' wallet when they click on an option from the dropdown.

### Why is this needed?

Different web3 apps run on different networks so it is convenient to display the current network to the user and letting them easily switch to the correct one.

### Proposed Solution or API

```tsx
import { NETWORKS, NetworkSwitch } from '@web3-ui/core';

const { mainnet, rinkeby, ropsten, polygon } = NETWORKS;

<NetworkSwitch networks=[mainnet, rinkeby, ropsten, polygon] />
```

### Additional Information

https://user-images.githubusercontent.com/39617427/149685304-76a3bf98-3920-4e54-a563-000085ed1b0c.mov

swellander:
Closes #247 

## Description

Adds a `<NetworkSwitch />` component to the core package.

## 📝 Additional Information
I wasn't sure of a clean way to integrate network logos with the [Chakra UI Select component](https://chakra-ui.com/docs/form/select) and could use some help with that if anyone has any ideas. Another option would be to make use of a library like [react-select](https://react-select.com/), which also has a [Chakra UI wrapper](https://www.npmjs.com/package/chakra-react-select). 

## Screenshot

https://user-images.githubusercontent.com/22231097/155431343-d63c30be-b602-4bda-b72e-7f3231b8dd48.mov

### DIFF
--- a/packages/core/src/components/NetworkSwitch/NetworkSwitch.tsx
+++ b/packages/core/src/components/NetworkSwitch/NetworkSwitch.tsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import { CHAIN_ID_TO_NETWORK } from '@web3-ui/hooks';
+import { Select, SelectProps } from '@chakra-ui/react';
+import { useChainId } from '@web3-ui/hooks/src/hooks';
+import { switchNetwork } from '@web3-ui/hooks/src/utils';
+
+export interface NetworkSwitchProps {
+  /**
+   * The chain IDs (in base 10) that the component should display as options
+   */
+  networks: number[];
+}
+
+/**
+ * A component to switch between a number of specified networks.
+ */
+export const NetworkSwitch: React.FC<NetworkSwitchProps & SelectProps> = ({
+  networks,
+  ...props
+}) => {
+  const chainId = useChainId();
+  const handleChange = (e: React.ChangeEvent<HTMLSelectElement>): void => {
+    const networkNum = Number(e.target.value);
+    switchNetwork(networkNum);
+  };
+  return (
+    <Select {...props} value={Number(chainId)} onChange={handleChange}>
+      {networks.map((network) => (
+        <option value={network}>{CHAIN_ID_TO_NETWORK[network]}</option>
+      ))}
+    </Select>
+  );
+};
--- a/.changeset/angry-pianos-destroy.md
+++ b/.changeset/angry-pianos-destroy.md
@@ -1,0 +1,6 @@
+---
+'@web3-ui/core': minor
+'@web3-ui/hooks': minor
+---
+
+Add `NetworkSwitch` component and `useChainId` hook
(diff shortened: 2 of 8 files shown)


packages/core/src/components/NetworkSwitch/NetworkSwitch.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { Select, SelectProps } from '@chakra-ui/react';
+
+const useChainId = () => {
+  const [chainId, setChainId] = useState<null | string>(null);
+  useEffect(() => {
+    window.ethereum.on('chainChanged', setChainId);
+    window.ethereum.request({ method: 'eth_chainId' }).then(setChainId);
+  }, []);
+  return chainId;
+};
```
swellander, line 12: Curious whether folks think we should just add this to the hooks package

packages/core/src/components/NetworkSwitch/NetworkSwitch.tsx
```diff
@@ -0,0 +1,59 @@
...
+        });
+      } catch (error) {
+        console.log(error);
+      }
+    } else {
+      alert(
+        'Switching networks automatically is only supported in MetaMask. Please consider installing it: https://metamask.io/download.html'
+      );
+    }
+  };
```
swellander, line 51: This `switchNetwork()` logic is also used in `useWallet()`. Would it make more sense to consolidate the two into a single function in a common util file?

Dhaiwat10: I think it would, yes

packages/core/src/components/NetworkSwitch/NetworkSwitch.tsx
Dhaiwat10: That's a great idea

Dhaiwat10: This only works with Metamask tho, right?

Dhaiwat10 (approved):
@swellander thanks for the PR. Can you please rebase this with `main` too?

Let's get this merged soon before we start re-working the entire project

packages/core/src/components/NetworkSwitch/NetworkSwitch.tsx
swellander: @Dhaiwat10 yeah, just MetaMask

swellander: Just reading your twitter post about the project's future. Should we then keep this `useChainId()` hook integrated with the `<NetworkSwitch/>` component, because we'll be pivoting away from hooks in general?

swellander: I went ahead and added it for now

PR #322, opened 2022-03-13
Switch to Radix, refactor all components

### LINKED ISSUE #321: Remove Chakra and setup Radix + Stitches
Recently I decided to get rid of the opinionated approach towards styling with this package and provide a blank slate to the user. A lot of early users of this package said that the fact that the package was so opinionated in terms of styling was a deal breaker.

This is why we will be replacing Chakra UI with [Radix](https://www.radix-ui.com/) and [Stitches](https://stitches.dev/).

Dhaiwat10:
Closes #321

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -19,8 +19,6 @@ module.exports = {
         ...config.resolve,
         alias: {
           ...config.resolve.alias,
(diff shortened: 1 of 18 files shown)


Canopix (approved):
Great Work!

PR #340, opened 2022-04-20
[hooks] Add `cacheProvider` prop to `<Provider />`

### LINKED ISSUE #310: Implement cached provider in Provider component
### Description

Currently connected state is not persisted, when I reload the page I have to connect to wallet again.

### Why is this needed?

I would like the feature to automatically connect to user wallet when user already connect it with the site once before already.

### Proposed Solution or API

when initiate web3modal we can pass cachedProvider flag

```
const web3Modal = new Web3Modal({
  ...
  cacheProvider: true,
});
```

when our page mounted we can check if there are cached provider

```
useEffect(() => {
  if (web3Modal.cachedProvider) {
    handleOnConnect();
  }
}, [])
```

### Additional Information

Example usage:
https://github.com/Web3Modal/web3modal/blob/master/example/src/App.tsx

jamescharlesworth:
See stories/useWriteContract for example.

Closes #310

## Description
This exposes cacheProvider as a prop to  `<Provider />` to allow persistence with Web3Modal.

## 📝 Additional Information

I'd like to implement some tests for this but wasn't sure how to mock accounts and didn't see any setup for that. Any ideas or examples on the best pattern to do that?

### DIFF
--- a/.changeset/strange-carrots-wonder.md
+++ b/.changeset/strange-carrots-wonder.md
@@ -1,0 +1,5 @@
+---
+'@web3-ui/hooks': minor
+---
+
+[hooks] `<Provider />` now exposes a `cacheProvider` prop to allow persistence with Web3Modal.
--- a/packages/hooks/src/Provider.tsx
+++ b/packages/hooks/src/Provider.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useRef } from 'react';
 import { JsonRpcSigner } from '@ethersproject/providers';
 import WalletConnectProvider from '@walletconnect/web3-provider';
 import { ethers } from 'ethers';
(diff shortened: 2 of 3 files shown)


RoseRompkxm:
This is exactly what this project is needed, why still no progress it's nearly 2 weeks!

Dhaiwat10:
LGTM, thanks @jamescharlesworth!

Dhaiwat10:
> I'd like to implement some tests for this but wasn't sure how to mock accounts and didn't see any setup for that. Any ideas or examples on the best pattern to do that?

I'm not really sure tbh. Maybe look into some existing tests that we have? I've seen some people use `ethereum-waffle` to mock stuff as well (that name might be slightly wrong)

Dhaiwat10:
@allcontributors please add @jamescharlesworth for code

